PR #11, opened 2024-03-15
chore: generated project structure using repo-config

sandovalrr:
- Generate project using repo-config
- Remove example files


.cookiecutter-replay.json
```diff
@@ -0,0 +1,46 @@
+{
+  "cookiecutter": {
+    "Introduction": "",
+    "type": "api",
```
cwasicki: This should be `lib` type. You can also compare here: https://github.com/frequenz-floss/frequenz-client-microgrid-python/blob/v0.x.x/.cookiecutter-replay.json

.cookiecutter-replay.json
```diff
@@ -0,0 +1,46 @@
...
+    "name": "frequenz-client-reporting-python",
```
cwasicki: Name would be `client-reporting` (according to the linked file).

proto/frequenz/api/frequenz_client_reporting_python/frequenz_client_reporting_python.proto
```diff
@@ -0,0 +1,23 @@
+// Frequenz Reporting API Client
```
cwasicki: This file should not be here.

py/frequenz/client/reporting/__init__.py
```diff
@@ -0,0 +1,7 @@
+# License: MIT
```
cwasicki: The whole `py` folder should also not be here.

cwasicki (changes requested):
There are some files related to API. Looks to me like you have to re-generate the whole repo with correct type and name and commit only that output.

sandovalrr:
> There are some files related to API. Looks to me like you have to re-generate the whole repo with correct type and name and commit only that output.

sure, let me do that

cwasicki:
@sandovalrr Can you confirm whether you committed the auto-generated files as-is?

PR #16, opened 2024-03-15
Introduce simple reporting client with support for single metric historical data

cwasicki:
This initial version of the reporting API client offers focused functionality to retrieve historical data for a single metric from a single component.

Features of this version:
* *Pagination Handling*: Seamlessly request time series data of a single metric from a single component, with integrated pagination management.
* *Data Transformation*: Utilize a wrapper data class that retains the raw protobuf response while offering transformation capabilities. A generator function is provided for iterating over individual page values.
* *Structured Data*: Streamline data representation through named tuples for timestamp and value pairs, eliminating ambiguity for single component and metric scenarios.
* *Usage Examples*: Code examples demonstrate usage, along with experimental code for upcoming features (considered for removal).
* *Unit Testing*: Basic unit tests are included.

Limitations of this version:
* *Single Metric Focus*: Initially supporting queries for individual metrics to support most common use cases, with an extensible design for multiple metrics/components in future.  -> 
  * https://github.com/frequenz-floss/frequenz-client-reporting-python/issues/19
* *States and bounds*:  In line with focus on widely used functionalities, the approach to integrating states and bounds within the data structures is still under exploration. -> 
  * https://github.com/frequenz-floss/frequenz-client-reporting-python/issues/20
  * https://github.com/frequenz-floss/frequenz-client-reporting-python/issues/25
* *Metric Sample Variants*: Currently supports `SimpleMetricSample` exclusively, with decision on how to integrate `AggregatedMetricSample` pending. -> 
  * https://github.com/frequenz-floss/frequenz-client-reporting-python/issues/21.
* *Resampling*: not yet exposed (service-side). -> 
  * https://github.com/frequenz-floss/frequenz-client-reporting-python/issues/22
* *Streaming*: Functions not yet available (service-side). Current generator implementations on pages and entries can be aligned with streaming output in future. -> 
  * https://github.com/frequenz-floss/frequenz-client-reporting-python/issues/23
* *Aggregation*: Support for formulas still missing (service-side) -> 
  * https://github.com/frequenz-floss/frequenz-client-reporting-python/issues/24

### DIFF
--- a/examples/client.py
+++ b/examples/client.py
@@ -1,0 +1,139 @@
+# License: MIT
+# Copyright © 2024 Frequenz Energy-as-a-Service GmbH
+
+"""Examples usage of reporting API."""
+
+import argparse
+import asyncio
+from datetime import datetime
+from pprint import pprint
+from typing import AsyncGenerator
+
+import pandas as pd
+from frequenz.client.common.metric import Metric
+
+from frequenz.client.reporting import ReportingClient
+
+# Experimental import
+from frequenz.client.reporting._client import MetricSample
+
+
+# pylint: disable=too-many-locals
+async def main(microgrid_id: int, component_id: int) -> None:
+    """Test the ReportingClient.
+
+    Args:
+        microgrid_id: int
+        component_id: int
+    """
+    service_address = "localhost:50051"
+    client = ReportingClient(service_address)
+
+    microgrid_components = [(microgrid_id, [component_id])]
+    metrics = [
+        Metric.DC_POWER,
+        Metric.DC_CURRENT,
+    ]
+
+    start_dt = datetime.fromisoformat("2023-11-21T12:00:00.00+00:00")
+    end_dt = datetime.fromisoformat("2023-11-21T12:01:00.00+00:00")
+
+    page_size = 10
+
+    print("########################################################")
+    print("Iterate over single metric generator")
+
+    async for sample in client.iterate_single_metric(
+        microgrid_id=microgrid_id,
+        component_id=component_id,
+        metric=metrics[0],
+        start_dt=start_dt,
+        end_dt=end_dt,
+        page_size=page_size,
+    ):
+        print("Received:", sample)
+
+    ###########################################################################
+    #
+    # The following code is experimental and demonstrates potential future
+    # usage of the ReportingClient.
+    #
+    ###########################################################################
+
+    async def components_data_iter() -> AsyncGenerator[MetricSample, None]:
+        """Iterate over components data.
+
+        Yields:
+            Single metric sample
+        """
+        # pylint: disable=protected-access
+        async for page in client._iterate_components_data_pages(
+            microgrid_components=microgrid_components,
+            metrics=metrics,
+            start_dt=start_dt,
+            end_dt=end_dt,
+            page_size=page_size,
+        ):
+            for entry in page.iterate_metric_samples():
+                yield entry
+
+    async def components_data_dict(
+        components_data_iter: AsyncGenerator[MetricSample, None]
+    ) -> dict[int, dict[int, dict[datetime, dict[Metric, float]]]]:
+        """Convert components data iterator into a single dict.
+
+            The nesting structure is:
+            {
+                microgrid_id: {
+                    component_id: {
+                        timestamp: {
+                            metric: value
+                        }
+                    }
+                }
+            }
+
+        Args:
+            components_data_iter: async generator
+
+        Returns:
+            Single dict with with all components data
+        """
+        ret: dict[int, dict[int, dict[datetime, dict[Metric, float]]]] = {}
+
+        async for ts, mid, cid, met, value in components_data_iter:
+            if mid not in ret:
+                ret[mid] = {}
+            if cid not in ret[mid]:
+                ret[mid][cid] = {}
+            if ts not in ret[mid][cid]:
+                ret[mid][cid][ts] = {}
+
+            ret[mid][cid][ts][met] = value
+
+        return ret
+
+    print("########################################################")
+    print("Iterate over generator")
+    async for msample in components_data_iter():
+        print("Received:", msample)
+
+    print("########################################################")
+    print("Dumping all data as a single dict")
+    dct = await components_data_dict(components_data_iter())
+    pprint(dct)
+
+    print("########################################################")
+    print("Turn data into a pandas DataFrame")
+    data = [cd async for cd in components_data_iter()]
+    df = pd.DataFrame(data).set_index("timestamp")
+    pprint(df)
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser()
+    parser.add_argument("microgrid_id", type=int, help="Microgrid ID")
+    parser.add_argument("component_id", type=int, help="Component ID")
+
+    args = parser.parse_args()
+    asyncio.run(main(args.microgrid_id, args.component_id))
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -29,6 +29,8 @@ requires-python = ">= 3.11, < 4"
 # TODO(cookiecutter): Remove and add more dependencies if appropriate
 dependencies = [
   "typing-extensions >= 4.5.0, < 5",
+  "frequenz-api-reporting >= 0.1.1, < 1",
+  "frequenz-client-common @ git+https://github.com/frequenz-floss/frequenz-client-common-python.git@v0.x.x",
 ]
 dynamic = ["version"]
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -64,6 +64,7 @@ dev-mypy = [
   "types-Markdown == 3.4.2.10",
   # For checking the noxfile, docs/ script, and tests
   "frequenz-client-reporting[dev-mkdocs,dev-noxfile,dev-pytest]",
+  "pandas-stubs >= 2, < 3",  # Only required for example
 ]
 dev-noxfile = [
   "nox == 2023.4.22",
@@ -73,6 +74,7 @@ dev-pylint = [
   "pylint == 3.0.2",
   # For checking the noxfile, docs/ script, and tests
   "frequenz-client-reporting[dev-mkdocs,dev-noxfile,dev-pytest]",
+  "pandas >= 2, < 3",  # Only required for example
 ]
 dev-pytest = [
   "pytest == 8.0.0",
@@ -83,6 +85,10 @@ dev-pytest = [
 ]
 dev = [
   "frequenz-client-reporting[dev-mkdocs,dev-flake8,dev-formatting,dev-mkdocs,dev-mypy,dev-noxfile,dev-pylint,dev-pytest]",
+]
+examples = [
+    "grpcio >= 1.51.1, < 2",
+    "pandas >= 2, < 3",
 ]
 
 [project.urls]
--- a/src/frequenz/client/reporting/_client.py
+++ b/src/frequenz/client/reporting/_client.py
@@ -1,0 +1,289 @@
+# License: MIT
+# Copyright © 2024 Frequenz Energy-as-a-Service GmbH
+
+"""Client for requests to the Reporting API."""
+
+from collections import namedtuple
+from dataclasses import dataclass
+from datetime import datetime
+from typing import Any, AsyncIterator, Awaitable, Generator, Type, cast
+
+import grpc.aio as grpcaio
+
+# pylint: disable=no-name-in-module
+from frequenz.api.common.v1.metrics.metric_sample_pb2 import Metric as PBMetric
+from frequenz.api.common.v1.microgrid.microgrid_pb2 import (
+    MicrogridComponentIDs as PBMicrogridComponentIDs,
+)
+from frequenz.api.common.v1.pagination.pagination_params_pb2 import (
+    PaginationParams as PBPaginationParams,
+)
+from frequenz.api.reporting.v1.reporting_pb2 import (
+    ListMicrogridComponentsDataRequest as PBListMicrogridComponentsDataRequest,
+)
+from frequenz.api.reporting.v1.reporting_pb2 import (
+    ListMicrogridComponentsDataResponse as PBListMicrogridComponentsDataResponse,
+)
+from frequenz.api.reporting.v1.reporting_pb2 import TimeFilter as PBTimeFilter
+from frequenz.api.reporting.v1.reporting_pb2_grpc import ReportingStub
+from frequenz.client.common.metric import Metric
+from google.protobuf.timestamp_pb2 import Timestamp as PBTimestamp
+
+# pylint: enable=no-name-in-module
+
+Sample = namedtuple("Sample", ["timestamp", "value"])
+"""Type for a sample of a time series."""
+
+MetricSample = namedtuple(
+    "MetricSample", ["timestamp", "microgrid_id", "component_id", "metric", "value"]
+)
+"""Type for a sample of a time series incl. metric type, microgrid and component ID"""
+
+
+@dataclass(frozen=True)
+class ComponentsDataPage:
+    """A page of microgrid components data returned by the Reporting service."""
+
+    _data_pb: PBListMicrogridComponentsDataResponse
+    """The underlying protobuf message."""
+
+    def is_empty(self) -> bool:
+        """Check if the page contains valid data.
+
+        Returns:
+            True if the page contains no valid data.
+        """
+        if not self._data_pb.microgrids:
+            return True
+        if not self._data_pb.microgrids[0].components:
+            return True
+        if not self._data_pb.microgrids[0].components[0].metric_samples:
+            return True
+        return False
+
+    def iterate_metric_samples(self) -> Generator[MetricSample, None, None]:
+        """Get generator that iterates over all values in the page.
+
+        Note: So far only `SimpleMetricSample` in the `MetricSampleVariant`
+        message is supported.
+
+
+        Yields:
+            A named tuple with the following fields:
+            * timestamp: The timestamp of the metric sample.
+            * microgrid_id: The microgrid ID.
+            * component_id: The component ID.
+            * metric: The metric name.
+            * value: The metric value.
+        """
+        data = self._data_pb
+        for mdata in data.microgrids:
+            mid = mdata.microgrid_id
+            for cdata in mdata.components:
+                cid = cdata.component_id
+                for msample in cdata.metric_samples:
+                    ts = msample.sampled_at.ToDatetime()
+                    met = Metric.from_proto(msample.metric).name
+                    value = (
+                        msample.sample.simple_metric.value
+                        if msample.sample.simple_metric
+                        else None
+                    )
+                    yield MetricSample(
+                        timestamp=ts,
+                        microgrid_id=mid,
+                        component_id=cid,
+                        metric=met,
+                        value=value,
+                    )
+
+    @property
+    def next_page_token(self) -> Any:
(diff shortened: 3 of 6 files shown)


pyproject.toml
```diff
@@ -29,6 +29,8 @@ requires-python = ">= 3.11, < 4"
 # TODO(cookiecutter): Remove and add more dependencies if appropriate
 dependencies = [
   "typing-extensions >= 4.5.0, < 5",
+  "frequenz-api-reporting >= 0.1.1, < 1",
+  "frequenz-client-common @ git+https://github.com/frequenz-floss/frequenz-client-common-python.git@v0.x.x",
```
cwasicki, line 33: This needs to be updated with the upcoming release.

flora-hofmann-frequenz (approved):
LGTM :+1:

examples/client.py
```diff
@@ -0,0 +1,139 @@
...
+        # pylint: disable=protected-access
+        async for page in client._iterate_components_data_pages(
+            microgrid_components=microgrid_components,
+            metrics=metrics,
+            start_dt=start_dt,
+            end_dt=end_dt,
+            page_size=page_size,
+        ):
+            for entry in page.iterate_metric_samples():
+                yield entry
```
tiyash-basu-frequenz, line 78: I am wondering if this helper function should exist in the library ..

src/frequenz/client/reporting/_client.py
```diff
@@ -0,0 +1,289 @@
...
+
+# pylint: enable=no-name-in-module
+
+Sample = namedtuple("Sample", ["timestamp", "value"])
+"""Type for a sample of a time series."""
+
+MetricSample = namedtuple(
+    "MetricSample", ["timestamp", "microgrid_id", "component_id", "metric", "value"]
+)
+"""Type for a sample of a time series incl. metric type, microgrid and component ID"""
```
tiyash-basu-frequenz, line 40: In theory, we could just use `MetricSample` everywhere, instead of using `Sample`. The higher level layers could create bespokse structs based on what they see fit. That would keep this library simpler, and in the future, we could just extend `MetricSample` to hold states, bounds, and aggregated metrics (instead of having to do that for `Sample` as well).

Any thoughts?

cwasicki: Yea, another way could be a `MetricSample` type from named tuple with all fields but timestamp and value optional. Depending on context the relevant fields would be set.

I would approach that in a next iteration where also the case of multiple metrics is implemented. As this is in a very early stage still, I think we can merge this for now.

tiyash-basu-frequenz (approved):
LGTM

src/frequenz/client/reporting/_client.py
llucax, line 40: Why not a dataclass? Performance reasons or just because it is more idiomatic for data science? If the reason is the former, it would be good to provide a benchmark to show the difference; if it is the later, it might be good to add a comment, and eventually explain what are the design principles for this library in the README or something, but for now, as we keep learning about what we really need, I would just put a comment in the code here.

src/frequenz/client/reporting/_client.py
```diff
@@ -0,0 +1,289 @@
...
+
+
+@dataclass(frozen=True)
+class ComponentsDataPage:
+    """A page of microgrid components data returned by the Reporting service."""
+
+    _data_pb: PBListMicrogridComponentsDataResponse
+    """The underlying protobuf message."""
```
llucax, line 48: I would call this `_pb_data` instead (or `_pb_message`). We call it `_raw` in the SDK. Maybe it would be nice to agree on some standard so all clients use the same, although it is not exposed so it is not that important.

src/frequenz/client/reporting/_client.py
```diff
@@ -0,0 +1,289 @@
...
+        """
+        if not self._data_pb.microgrids:
+            return True
+        if not self._data_pb.microgrids[0].components:
+            return True
+        if not self._data_pb.microgrids[0].components[0].metric_samples:
+            return True
+        return False
+
+    def iterate_metric_samples(self) -> Generator[MetricSample, None, None]:
```
llucax, line 64: I would just make the class iterable directly (also `Iterator[MetricSample]`[^1] is a shorter and more idiomatic alias for `Generator[MetricSample, None, None]`[^2])

```py
    def __iter__(self) -> Iterator[MetricSample]:
        ...
```

Then you can just do `for sample in page:` instead of `for sample in page.iterate_metric_samples():`.

[^1]: Technically a [`Generator`](https://peps.python.org/pep-0342/) is more than an `Iterator`:

    > extends [iterators](https://docs.python.org/3.11/glossary.html#term-iterator) with the [send()](https://docs.python.org/3.11/reference/expressions.html#generator.send), [throw()](https://docs.python.org/3.11/reference/expressions.html#generator.throw) and [close()](https://docs.python.org/3.11/reference/expressions.html#generator.close) methods

    But in this context I think we only care about the iterator part, we are using a generator out of convenience, not because we are actually using the extra methods provided by a generator.

[^2]: From [`typing.Generator`](https://docs.python.org/3.11/library/typing.html#typing.Generator):

    > Alternatively, annotate your generator as having a return type of either Iterable[YieldType] or Iterator[YieldType]:

src/frequenz/client/reporting/_client.py
```diff
@@ -0,0 +1,289 @@
...
+                    yield MetricSample(
+                        timestamp=ts,
+                        microgrid_id=mid,
+                        component_id=cid,
+                        metric=met,
+                        value=value,
+                    )
+
+    @property
+    def next_page_token(self) -> Any:
```
llucax, line 101: This should return `str`, right? If you want it to return an *opaque* type, then use `object` instead, as `Any` will completely disable type checking, doing wrong stuff like `t = p.next_page_token; t.i_dont_exist()` will not be detected by `mypy`.

src/frequenz/client/reporting/_client.py
```diff
@@ -0,0 +1,289 @@
...
+        """Get the token for the next page of data.
+
+        Returns:
+            The token for the next page of data.
+        """
+        return self._data_pb.pagination_info.next_page_token
+
+
+class ReportingClient:
```
llucax, line 110: We should probably also standarize this name scheme. I think in the microgrid API I used just `ApiClient`, as it is expected to be used as `microgrid.ApiClient`. I don't mind using `xxx.ApiClient` or `XxxApiClient` but I think we should call it `ApiClient` because otherwise it might be confusing for some APIs if the term *client* have some other meaning.

src/frequenz/client/reporting/_client.py
```diff
@@ -0,0 +1,289 @@
...
+        """Create a new Reporting client.
+
+        Args:
+            service_address: The address of the Reporting service.
+        """
+        self._grpc_channel = grpcaio.insecure_channel(service_address)
+        self._stub = ReportingStub(self._grpc_channel)
+
+    # pylint: disable=too-many-arguments
+    async def iterate_single_metric(
```
llucax, line 123: I wouldn't name these methods `iterate`, not sure if they are a common idiom in data science but otherwise I'd say it is not a common idiom in general in python. I think client methods should map more or less one to one to the gRPC API methods. In this case I think the name should be: `list_microgrid_components_data()`. I think it is fine to do the page unwrapping and still use the same name as the gRPC API method.

Again, I think we should have a common approach in all APIs.

PS: I guess the extra explicitness in names might come from *untyped python*, in that case it makes more sense, but when using type hints you can already see you are iterating (`AsyncIterator`) over a single metric (`[Sample]`).

src/frequenz/client/reporting/_client.py
```diff
@@ -0,0 +1,289 @@
...
+            metrics=[metric],
+            start_dt=start_dt,
+            end_dt=end_dt,
+            page_size=page_size,
+        ):
+            for entry in page.iterate_metric_samples():
+                yield Sample(timestamp=entry.timestamp, value=entry.value)
+
+    # pylint: disable=too-many-arguments
+    async def _iterate_components_data_pages(
```
llucax, line 159: I would probably expose this publicly, in case someone wants to get pages for some reason, I don't see a reason to hide it.

Even more, now with the code in front of my eyes, I think the best approach would be to `list_microgrid_components_data()` return an object with a few iterable properties:

```py
async for page in client.list_microgrid_components_data(...).pages:
    ...

async for sample in client.list_microgrid_components_data(...).samples:
    ...
```

So `list_microgrid_components_data()` could return a `Response` object with properties like (pseudo-code using the current method names):

```py
@property
async def samples(self) -> AsyncIterator[Sample]:
    return self.client._iterate_single_metric(...)

@property
async def pages(self) -> AsyncIterator[ComponentsDataPage]:
    return self.client._iterate_components_data_pages(...)
```

This design might need a bit more thought, so maybe we should wait on it and discuss it in a separate issue.

src/frequenz/client/reporting/_client.py
```diff
@@ -0,0 +1,289 @@
...
+        Args:
+            _exc_type: Type of exception raised in the async context.
+            _exc_val: Exception instance raised.
+            _exc_tb: Traceback object at the point where the exception occurred.
+
+        Returns:
+            None, allowing any exceptions to propagate.
+        """
+        await self.close()
+        return None
```
llucax, line 289: This might be a good foundation for a `GrpcApiClient` class that can live in https://github.com/frequenz-floss/frequenz-client-base-python/!

It could also standarize the constructor, I think right now some clients take a grpc channel and some take a connection string. I think we should go with the later to remove the explicit dependency on grpc for downstream projects.

llucax (commented):
I arrived quite late to the party, sorry, but it is actually a good think, I wouldn't have wanted to block the PR with this feedback, at least some of it is more long term.

I will create separate issues for things that need further discussion.

src/frequenz/client/reporting/_client.py
llucax: This one I think should be answered here as it is not worth a separate issue, unless this triggers a bigger discussion.

src/frequenz/client/reporting/_client.py
llucax: * https://github.com/frequenz-floss/frequenz-api-common/discussions/215

src/frequenz/client/reporting/_client.py
llucax: This is a small change, so we can also briefly discuss here and could be quickly changed in the current repo. The broader discussion is now here:

* https://github.com/frequenz-floss/frequenz-client-base-python/discussions/41

src/frequenz/client/reporting/_client.py
llucax: I suggest also discussing this one here, and only create a separate issue if it is not trivial.

src/frequenz/client/reporting/_client.py
llucax: IMHO it is pretty clear that Python API client method names should match the gRPC API, so I suggest to create a separate issue about this only if anyway thinks differently.

As for how to iterate, please refer to:

* https://github.com/frequenz-floss/frequenz-client-base-python/discussions/41

src/frequenz/client/reporting/_client.py
llucax: * https://github.com/frequenz-floss/frequenz-client-base-python/discussions/41

src/frequenz/client/reporting/_client.py
llucax: * https://github.com/frequenz-floss/frequenz-client-base-python/issues/42

src/frequenz/client/reporting/_client.py
llucax: I suggest discussing this in a meeting, I guess it should be quicker in a more interactive way.

src/frequenz/client/reporting/_client.py
cwasicki: The idea was to keep it as low-level as possible, since this is the fundamental data structure that is repeated for every single sample. So for me the question is, why should we use a dataclass instead? For further analysis the data is typically transformed into an array-like (e.g. pandas data frame) data type anyway, which nicely works with this data type (originally it was just a tuple until @shsms passed by).

src/frequenz/client/reporting/_client.py
cwasicki: Good idea to standardize, I am fine with all suggestions.

src/frequenz/client/reporting/_client.py
cwasicki: You are right, should be string or None: https://github.com/frequenz-floss/frequenz-api-common/blob/20cdc478555f3ba55bd86ccfd19dbc0beaf80edc/proto/frequenz/api/common/v1/pagination/pagination_info.proto#L26.

src/frequenz/client/reporting/_client.py
cwasicki: >I wouldn't name these methods iterate,

> PS: I guess the extra explicitness in names might come from untyped python, in that case it makes more sense, but when using type hints you can already see you are iterating (AsyncIterator) over a single metric ([Sample]).

Agree, can remove this. Actually the explicitness is to distinguish from different access methods that I considered, e.g. alternative to the iterator a dictionary representing all nesting levels. At the moment I am not sure whether it will come.

> In this case I think the name should be: list_microgrid_components_data()

Not sure about this. The method offers only a small subset of the potential functionality of the gRPC function. And the  current name reflects this limitation. Also, since the current version is sufficient for many applications already, I am not sure whether this method will be extended or another method will be added, that provides full support of the gRPC.

src/frequenz/client/reporting/_client.py
cwasicki: > I would probably expose this publicly, in case someone wants to get pages for some reason, I don't see a reason to hide it.

Honestly I don't see a reason to expose it now. If we realize this becomes important, it's easy to expose. Because I don't see a use-case where you would want the pages that you cannot achieve with the iterator. 

> Even more, now with the code in front of my eyes, I think the best approach would be to list_microgrid_components_data()

That could be an option. But do you have a good example where having pages are better than an iterator over the samples?

src/frequenz/client/reporting/_client.py
cwasicki: Sounds good

src/frequenz/client/reporting/_client.py
cwasicki: >I would just make the class iterable directly (also Iterator[MetricSample][1](https://github.com/frequenz-floss/frequenz-client-reporting-python/pull/16/files#user-content-fn-1-bbcbfbb94d5c423a8a52d25024597561) is a shorter and more idiomatic alias for Generator[MetricSample, None, None][2](https://github.com/frequenz-floss/frequenz-client-reporting-python/pull/16/files#user-content-fn-2-bbcbfbb94d5c423a8a52d25024597561))

Agree, will change this. In general a generator was used for lazy iteration, but that doesn't need to be reflected in the type annotation.

src/frequenz/client/reporting/_client.py
llucax: Yeah, agreed, that was my main question. As for why dataclasses, just because they are a higher level abstraction, and are more flexible, but if this is targeting data science and for that use case tuples make more sense, then I agree we should use that. I just think we should make it more clear, as said above, a comment in the code should be enough.

src/frequenz/client/reporting/_client.py
llucax: Iterators also provides *lazy iteration*. Generators are just more powerful as they also let you pass messages between the iterator and the outside world.

src/frequenz/client/reporting/_client.py
llucax: OK, should we create an issue or do you plan to do a quick fix directly?

src/frequenz/client/reporting/_client.py
llucax: Mmmm, OK, I see. In general I think clients should only be a thin wrapper over the gRPC calls, and other more convenient functions could be built on top. I mean, I'm completely fine to keep it as it is for an early version, especially if we need something fast, but in the long term I don't think it is the right approach. If the API is more flexible, then some users must need that flexibility, otherwise the gRPC API we have is more complex than needed.

src/frequenz/client/reporting/_client.py
llucax: > Because I don't see a use-case where you would want the pages that you cannot achieve with the iterator.

Performance. We tested this and iterating over items was considerably slower than going through pages.

llucax: Again, maybe because of data is used when doing data science, at the end is the same for that use case, but I think the API client should be pretty flexible and general purpose. And again, I'm thinking more long term, not just about what we need right now. For the right now I agree it is fine and we can improve later, but if we agree about this, I would create issues about it so:
1. We don't forget
2. It is clear for everyone what's the future direction we want to take, so issues with that future direction can also be raised early if anyone sees any.

src/frequenz/client/reporting/_client.py
cwasicki: https://github.com/frequenz-floss/frequenz-client-reporting-python/pull/28

src/frequenz/client/reporting/_client.py
cwasicki: Decision was to use `XxxApiClient`, right?

cwasicki: https://github.com/frequenz-floss/frequenz-client-reporting-python/pull/29

src/frequenz/client/reporting/_client.py
cwasicki: https://github.com/frequenz-floss/frequenz-client-reporting-python/issues/30

src/frequenz/client/reporting/_client.py
cwasicki: https://github.com/frequenz-floss/frequenz-client-reporting-python/issues/31

src/frequenz/client/reporting/_client.py
cwasicki: >If the API is more flexible, then some users must need that flexibility, otherwise the gRPC API we have is more complex than needed.

Yes, the flexibility of the API will be exposed, there are issues for all of the missing features. Question is whether we have only one method that exposes all or multiple. I would postpone the decision until we address some of the open issues, e.g. https://github.com/frequenz-floss/frequenz-client-reporting-python/issues/19.

src/frequenz/client/reporting/_client.py
cwasicki: https://github.com/frequenz-floss/frequenz-client-reporting-python/issues/32

src/frequenz/client/reporting/_client.py
cwasicki: Partly addressed here: https://github.com/frequenz-floss/frequenz-client-reporting-python/pull/33.

On whether to keep the existing single component function, see https://github.com/frequenz-floss/frequenz-client-reporting-python/issues/34.

On the iterator, see https://github.com/frequenz-floss/frequenz-client-reporting-python/issues/31.

src/frequenz/client/reporting/_client.py
cwasicki: https://github.com/frequenz-floss/frequenz-client-reporting-python/pull/35

src/frequenz/client/reporting/_client.py
cwasicki: https://github.com/frequenz-floss/frequenz-client-reporting-python/pull/36

PR #26, opened 2024-03-22
Add support for requesting multiple metrics

cwasicki:
This add support to request multiple metrics in a single request. The limitation to query data only from a single component is still in place. Moreover the distinct return types when querying single or multiple metrics is removed.

In addition to that, this turns the example client into a command line tool that can be used to extract data from the reporting API and print it to stdout. Besides option for selecting the data, the service address and the display format can be changed via command line.


tiyash-basu-frequenz (approved):
LGTM! Thanks!

PR #28, opened 2024-04-12
Fix return type for next_page_token method

cwasicki:
From https://github.com/frequenz-floss/frequenz-client-reporting-python/pull/16#discussion_r1548118756


flora-hofmann-frequenz (approved):
LGTM!

PR #33, opened 2024-04-12
Remove iterator from function names

cwasicki:
Replace iterator in method names to be closer to gRPC names. 

Raised here https://github.com/frequenz-floss/frequenz-client-reporting-python/pull/16/files#r1548128616.

Naming might not be final.

FYI @llucax


flora-hofmann-frequenz (approved):
:+1:

Issue #19, opened 2024-03-21
Support multiple components and microgrids


cwasicki:
### What's needed?

In the initial version only a combination of one metric, component and microgrid ID each can be requested. The API supports arbitrary lists of microgrid-component IDs and metric IDs. This  functionality should be exposed.

In a first step, multiple metrics for a single component is implemented in https://github.com/frequenz-floss/frequenz-client-reporting-python/pull/26. Multiple components will be addressed separately.

### Proposed solution

The case of multiple metrics, components has been considered in the current design already and only needs to be exposed. First steps in this direction are addressed for multiple metrics in https://github.com/frequenz-floss/frequenz-client-reporting-python/pull/26.

### Additional context

The API requests the cross-combination of the elements in the list of `microgrid_components` and the elements in the list of `metrics`, which might not provide sufficient control over the combinations (e.g. too many or non-existing). For instance
```
metrics = [SOC, power]
microgrid_components = [ {MID, CID_meter}, {MID, CID_battery} ]
```
leads to 4 combinations.

_EDIT: For now we focus on exposing the signature as close to the underlying  API as possible and keep the following as potential future enhancements:_

Alternatively we could require the user to provide request data in the format
```
[[mid1, cid1, metric1], [mid2, cid2, metric2], ...]
```
and potentially have multiple requests whose output needs to be merged.

cwasicki:
Updated scope of this PR since multiple metrics are now addressed here: https://github.com/frequenz-floss/frequenz-client-reporting-python/pull/26.

cwasicki:
Updated issue description narrowing down the scope.

flora-hofmann-frequenz:
Closing issue after merging PR #39.

PR #39, opened 2024-04-19
Add support for multiple components and microgrids

flora-hofmann-frequenz:
A simple support for mutliple components and microgrids addressing Issue #19.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -10,5 +10,9 @@
 
 ## New Features
 
+* The client method `list_multiple_microgrid_components_data` is added to
+expose the functionality of supporting arbitrary lists of microgrid-component IDs
+and metric IDs. It is intended to be as close as possible on the gRPC function call.
+
 ## Bug Fixes
 
--- a/src/frequenz/client/reporting/_client.py
+++ b/src/frequenz/client/reporting/_client.py
@@ -158,6 +158,44 @@ class ReportingApiClient:
                 yield entry
 
     # pylint: disable=too-many-arguments
+    async def list_multiple_microgrid_components_data(
+        self,
+        *,
+        microgrid_components: list[tuple[int, list[int]]],
+        metrics: Metric | list[Metric],
+        start_dt: datetime,
+        end_dt: datetime,
+        page_size: int = 1000,
+    ) -> AsyncIterator[MetricSample]:
+        """Iterate over the data for multiple microgrids and components.
+
+        Args:
+            microgrid_components: List of tuples where each tuple contains
+                                  microgrid ID and corresponding component IDs.
+            metrics: The metric name or list of metric names.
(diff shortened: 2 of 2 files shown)


RELEASE_NOTES.md
```diff
@@ -4,9 +4,7 @@
 
 ## Upgrading
 
-* The `ReportingClient` is renamed to `ReportingApiClient`.
-
-* The client method `iterate_single_component` is renamed to `list_single_component_data`.
+* The client method `list_multiple_microgrid_components_data` is added.
```
cwasicki: * The exsisting upgrading hints should not be removed.
* Maybe add a bit more context that this is intended to be as close as possible on the gRPC function call.
* This is a new feature and not relevant for people upgrading to the next version.

src/frequenz/client/reporting/_client.py
```diff
@@ -157,6 +157,44 @@ async def list_single_component_data(
...
+            start_dt: The start date and time.
+            end_dt: The end date and time.
+            page_size: The page size.
+
+        Yields:
+            A named tuple with the following fields:
+            * timestamp: The timestamp of the metric sample.
+            * value: The metric value.
+        """
+        async for page in self._list_microgrid_components_data_pages(
```
cwasicki, line 188: Since we should be as close as possible to the original function, we should use the same arguments as for `_list_microgrid_components_data_pages`

cwasicki (changes requested):

src/frequenz/client/reporting/_client.py
flora-hofmann-frequenz: Meaning that I should use `microgrid_components: list[tuple[int, list[int]]]` instead of `microgrid_ids: list[int], component_ids: list[int]`?

cwasicki: Yes, because that reflects the data structure used in the underlying gRPC function.

src/frequenz/client/reporting/_client.py
```diff
@@ -157,6 +157,41 @@ async def list_single_component_data(
...
+                                  microgrid ID and corresponding component IDs.
+            metrics: The metric name or list of metric names.
+            start_dt: The start date and time.
+            end_dt: The end date and time.
+            page_size: The page size.
+
+        Yields:
+            A named tuple with the following fields:
+            * timestamp: The timestamp of the metric sample.
+            * value: The metric value.
```
cwasicki, line 186: The documentation should be like [here](https://github.com/frequenz-floss/frequenz-client-reporting-python/pull/39/files#diff-bc44ef8bfdafafc5ae638f3475097e7459739c5eda848f6954aa620b48f2e50aL73)

cwasicki (changes requested):
LGTM, just minor doc comment

cwasicki (approved):

PR #42, opened 2024-04-24
clear release notes for preparetion for v0.2.0 release

mhrrmk:


cwasicki (changes requested):
The release notes should only be cleared after the release.

Issue #22, opened 2024-03-21
Support for resampling option


cwasicki:
### What's needed?

The resampling should be exposed if it is supported by the service.

While the client should support any resolution, it needs to be decided whether the service supports these too. If the service only supports a pre-defined set of resolutions, the client has to be extended with resampling capabilities. 

In this case, exposing the service resampling and the client implementation can be done in two iterations.

### Additional context

https://github.com/frequenz-floss/frequenz-api-reporting/issues/19

flora-hofmann-frequenz:
PR #52 merged, thus close this issue.

PR #52, opened 2024-05-03
Expose resampling option

flora-hofmann-frequenz:
Addressing issue #22 by exposing the resampling option as an optional input.

### DIFF
--- a/examples/client.py
+++ b/examples/client.py
@@ -49,6 +49,7 @@ def main() -> None:
         help="End datetime in YYYY-MM-DDTHH:MM:SS format",
         required=True,
     )
+    parser.add_argument("--resolution", type=int, help="Resolution", default=None)
     parser.add_argument("--psize", type=int, help="Page size", default=100)
     parser.add_argument(
         "--display", choices=["iter", "df", "dict"], help="Display format", default="df"
@@ -61,6 +62,7 @@ def main() -> None:
             args.metrics,
             args.start,
             args.end,
+            args.resolution,
             page_size=args.psize,
             service_address=args.url,
             display=args.display,
@@ -68,13 +70,14 @@ def main() -> None:
     )
 
 
-# pylint: disable=too-many-arguments
+# pylint: disable=too-many-arguments, too-many-locals
 async def run(
     microgrid_id: int,
     component_id: int,
     metric_names: list[str],
     start_dt: datetime,
     end_dt: datetime,
+    resolution: int,
     page_size: int,
     service_address: str,
     display: str,
@@ -87,6 +90,7 @@ async def run(
         metric_names: list of metric names
         start_dt: start datetime
         end_dt: end datetime
+        resolution: resampling resolution in sec
         page_size: page size
         service_address: service address
         display: display format
@@ -112,6 +116,7 @@ async def run(
             metrics=metrics,
             start_dt=start_dt,
             end_dt=end_dt,
+            resolution=resolution,
             page_size=page_size,
         )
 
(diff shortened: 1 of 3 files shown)


examples/client.py
```diff
@@ -49,6 +49,7 @@ def main() -> None:
         help="End datetime in YYYY-MM-DDTHH:MM:SS format",
         required=True,
     )
+    parser.add_argument("--resolution", type=int, help="Resolution", default=1)
```
cwasicki: That can also be None, i.e. no resampling. I would also make this the default.

src/frequenz/client/reporting/_client.py
```diff
@@ -130,6 +133,7 @@ async def list_single_component_data(
         metrics: Metric | list[Metric],
         start_dt: datetime,
         end_dt: datetime,
+        resolution: int = 1,
```
cwasicki: Also here it can be `None` too, and I suggest to make this the default.

src/frequenz/client/reporting/_client.py
```diff
@@ -165,6 +171,7 @@ async def list_microgrid_components_data(
...
```
cwasicki: See above

src/frequenz/client/reporting/_client.py
```diff
@@ -203,6 +212,7 @@ async def _list_microgrid_components_data_pages(
         metrics: list[Metric],
         start_dt: datetime,
         end_dt: datetime,
+        resolution: int = 1,
```
cwasicki: I would remove the default here completely. (And for page size too IMO)

cwasicki (commented):
Just minor comments on default, otherwise LGTM.

Release notes are still missing.

src/frequenz/client/reporting/_client.py
```diff
@@ -235,8 +246,13 @@ def dt2ts(dt: datetime) -> PBTimestamp:
             end=dt2ts(end_dt),
         )
 
+        resampling_options = PBResamplingOptions(
+            resolution=resolution,
+        )
```
cwasicki: You could move this into [this line](https://github.com/frequenz-floss/frequenz-client-reporting-python/pull/52/files#diff-bc44ef8bfdafafc5ae638f3475097e7459739c5eda848f6954aa620b48f2e50aR255) to avoid the CI failure.

src/frequenz/client/reporting/_client.py
```diff
@@ -130,6 +133,7 @@ async def list_single_component_data(
         metrics: Metric | list[Metric],
         start_dt: datetime,
         end_dt: datetime,
+        resolution: Optional[int] = None,
```
cwasicki: You don't need the `Optional`, this can be `int | None`

src/frequenz/client/reporting/_client.py
```diff
@@ -6,7 +6,7 @@
 from collections import namedtuple
 from dataclasses import dataclass
 from datetime import datetime
-from typing import Any, AsyncIterator, Awaitable, Iterator, Type, cast
+from typing import Any, AsyncIterator, Awaitable, Iterator, Optional, Type, cast
```
cwasicki: The `Optional` can go.

src/frequenz/client/reporting/_client.py
```diff
@@ -203,7 +212,8 @@ async def _list_microgrid_components_data_pages(
         metrics: list[Metric],
         start_dt: datetime,
         end_dt: datetime,
-        page_size: int = 1000,
+        resolution: Optional[int],
```
cwasicki: One more `Optional` left.

cwasicki (changes requested):

src/frequenz/client/reporting/_client.py
flora-hofmann-frequenz: When I took it out yesterday, I got an error here: `return self._data_pb.pagination_info.next_page_token`

Now I tried it again, and the error's not there anymore? Sorry about that!
I have taken it out.

cwasicki (approved):

PR #53, opened 2024-05-03
Update README with usage instructions

cwasicki:
Simple usage example to query multiple metrics from single and multiple microgrids and components.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,3 +20,88 @@ The following platforms are officially supported (tested):
 
 If you want to know how to build this project and contribute to it, please
 check out the [Contributing Guide](CONTRIBUTING.md).
+
+
+## Usage
+
+Please also refer to [examples](https://github.com/frequenz-floss/frequenz-client-reporting-python/tree/HEAD/examples) for more detailed usage.
+
+### Installation
+
+```bash
+# Choose the version to install
+VERSION=0.2.0
+pip install frequenz-client-reporting==$VERSION
+```
+
+
+### Initialize the client
+
+```python
+from datetime import datetime
+
+from frequenz.client.common.metric import Metric
+from frequenz.client.reporting import ReportingApiClient
+
+# Change server address
+SERVICE_ADDRESS = "localhost:4711"
+client = ReportingApiClient(service_address=SERVICE_ADDRESS)
+```
+
+### Query metrics for a single microgrid and component:
+
+```python
+data = [
+    sample async for sample in
+    client.list_single_component_data(
+        microgrid_id=1,
+        component_id=100,
+        metrics=[Metric.AC_ACTIVE_POWER, Metric.AC_REACTIVE_POWER],
+        start_dt=datetime.fromisoformat("2024-05-01T00:00:00"),
+        end_dt=datetime.fromisoformat("2024-05-02T00:00:00"),
+        page_size=10000,
+    )
+]
+```
+
+
+### Query metrics for multiple microgrids and components
+
+```python
+# Set the microgrid ID and the component IDs that belong to the microgrid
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -20,3 +20,77 @@ The following platforms are officially supported (tested):
 
 If you want to know how to build this project and contribute to it, please
 check out the [Contributing Guide](CONTRIBUTING.md).
+
+
+## Usage
+
+Please also refer to [examples](examples) for more detailed usage.
```
llucax: The readme is included as-is as the [home page for the generated docs](https://github.com/frequenz-floss/frequenz-client-reporting-python/blob/v0.x.x/docs/index.md). This link won't work in the generated docs unless you also add some `examples/` section to the `docs/`.

Another alternative is to split the README into *includable* sections, and then in the generated docs home page, only include the sections that makes sense, like we do in the channels repo:
* https://raw.githubusercontent.com/frequenz-floss/frequenz-channels-python/v1.x.x/README.md
* https://raw.githubusercontent.com/frequenz-floss/frequenz-channels-python/v1.x.x/docs/index.md

You need to install the [`mkdocs-include-markdown-plugin`](https://pypi.org/project/mkdocs-include-markdown-plugin/) for this to work.

README.md
```diff
@@ -20,3 +20,77 @@ The following platforms are officially supported (tested):
...
+### Installation
+
+```bash
+# Choose the version to install
+VERSION=0.2.0
+pip install frequenz-client-reporting==$VERSION
+```
+
+
+### Initialize the client
```
llucax, line 38: Nitpick: We usually leave an empty line after headings:

```suggestion
### Initialize the client

```

Same for the other headings...

README.md
cwasicki: >This link won't work in the generated docs unless you also add some examples/ section to the docs/.

Do you mean a subfolder? I have specified the full URL now.

cwasicki:
Updated and added instructions how to use the command line client example.

README.md
leandro-lucarella-frequenz: Yes and no, now the link will still work in the generated docs but IMHO is not ideal. It will point to the HEAD, which might be different than the version you are looking the docs for, for example, which could lead to confusion for users. You could get a "dynamic" link using macros, but if you do so, then the README will be screwed when read directly from GitHub.

What I mean is, again, ideally, is what I show above we did with the channels repo, where we include only some parts of the README in the generated docs.

A separate, but interesting discussion, is if we want to have an "examples" section in the generated docs, and have also kind of pick up the examples in `examples/` automatically or something. This could be done with some magic.

All that said, I'm OK with the current state as a temporary solution, just saying it has its flaws.

llucax: Yes and no, now the link will still work in the generated docs but IMHO is not ideal. It will point to the HEAD, which might be different than the version you are looking the docs for, for example, which could lead to confusion for users. You could get a "dynamic" link using macros, but if you do so, then the README will be screwed when read directly from GitHub.

What I mean is, again, ideally, is what I show above we did with the channels repo, where we include only some parts of the README in the generated docs.

A separate, but interesting discussion, is if we want to have an "examples" section in the generated docs, and have also kind of pick up the examples in `examples/` automatically or something. This could be done with some magic.

All that said, I'm OK with the current state as a temporary solution, just saying it has its flaws.

cwasicki: Ok, thank you for explaining. I would leave this for a future iteration.

tiyash-basu-frequenz (approved):

tiyash-basu-frequenz:
@cwasicki the commit needs to be signed.

Issue #41, opened 2024-04-22
Update to reporting api v0.2.0


cwasicki:
Update reporting API dependency to v0.2.0: https://github.com/frequenz-floss/frequenz-api-reporting/releases/tag/v0.2.0

flora-hofmann-frequenz:
It looks like I first need to update the common release, otherwise I get nox errors...

PR #58, opened 2024-05-15
Update reporting api dependency

flora-hofmann-frequenz:
Addressing issue #41

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -6,11 +6,13 @@
 
 ## Upgrading
 
-<!-- Here goes notes on how to upgrade from previous versions, including deprecations and what they should be replaced with -->
+* Upgrading "frequenz-api-reporting >= 0.1.1, < 0.2" to "frequenz-api-reporting >= 0.2.0, < 0.3"
+* Upgrading "frequenz-client-common == 0.1.0" to "frequenz-client-common >= 0.2.0, < 0.3"
 
 ## New Features
 
-<!-- Here goes the main new features and examples or instructions on how to use them -->
+* The `ReportingApiClient` exposes the resampling option of the data, with its 
+resolution represented in second and its default set to `None`.
 
 ## Bug Fixes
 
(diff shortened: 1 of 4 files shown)


RELEASE_NOTES.md
```diff
@@ -6,11 +6,13 @@
...
 
-<!-- Here goes notes on how to upgrade from previous versions, including deprecations and what they should be replaced with -->
+* Upgrading "frequenz-api-reporting >= 0.1.1, < 0.2" to "frequenz-api-reporting >= 0.2.0, < 0.3"
+* Upgrading "frequenz-client-common == 0.1.0" to "frequenz-client-common >= 0.2.0, < 0.3"
 
 ## New Features
 
-<!-- Here goes the main new features and examples or instructions on how to use them -->
+* The `ReportingApiClient` exposes the resampling option of the data, with its 
+resolution represented in second and its default set to `None`.
```
cwasicki, line 15: This is not related to this update, should it be here? If so, better make this a separate commit.

docs/_scripts/macros.py
```diff
@@ -30,7 +30,7 @@ def _slugify(text: str) -> str:
     # The type of the return value is not defined for the markdown library.
     # Also for some reason `mypy` thinks the `toc` module doesn't have a
     # `slugify_unicode` function, but it definitely does.
-    return toc.slugify_unicode(text, "-")  # type: ignore[attr-defined,no-any-return]
+    return toc.slugify_unicode(text, "-")
```
cwasicki, line 33: Also a separate commit.

pyproject.toml
cwasicki, line 1: Are all of these updates in this file needed for the reporting API update?

RELEASE_NOTES.md
flora-hofmann-frequenz: I thought so too, but then it does not appear in our merged code base which is weird?
I had separate commits but squashed them - will add them again.

pyproject.toml
flora-hofmann-frequenz: Yes. The toml was quite outdated/didn't have all the dependencies needed and caused problems with the dependencies (I discussed this with Luca [here](https://frequenz-io.slack.com/archives/CNMCCUB5H/p1715934534748859)). All these updates are needed for the reporting API dependency update (that's also why the dependabot update failed).

cwasicki (approved):

PR #59, opened 2024-05-21
Include states in data structure

flora-hofmann-frequenz:
To address [this issue](https://github.com/frequenz-floss/frequenz-client-common-python/issues/29).

The structure should be as in protobuf definition.

### DIFF
--- a/src/frequenz/client/reporting/_client.py
+++ b/src/frequenz/client/reporting/_client.py
@@ -66,41 +66,50 @@ class ComponentsDataPage:
             return True
         return False
 
-    def __iter__(self) -> Iterator[MetricSample]:
+    def __iter__(self) -> Iterator[dict[Any, Any]]:
         """Get generator that iterates over all values in the page.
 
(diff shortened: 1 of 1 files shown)


flora-hofmann-frequenz:
I am currently getting empty `states []` outputs, would I always get an empty state unless there is a state message, error or warning?

src/frequenz/client/reporting/_client.py
```diff
@@ -66,41 +66,50 @@ def is_empty(self) -> bool:
             return True
         return False
 
-    def __iter__(self) -> Iterator[MetricSample]:
+    def __iter__(self) -> Iterator[dict[Any, Any]]:
```
cwasicki, line 69: Sorry, I think there was a misunderstanding. The idea was to have the protobuf message represented by a dedicated dataclass. A dict is flexible but wouldn't help us for type safety, which is important here.

flora-hofmann-frequenz:
Closing this as it will be on-hold until there is a use case with states.

PR #67, opened 2024-06-11
Use streaming endpoint in list_microgrid_components_data

cwasicki:
The streaming endpoint supports requests for historical data. This replaces the list endpoint with the streaming endpoint. The names in the client are for now unchanged to not break user code. They might be revised when live streaming will also be supported by the service.

The list endpoint will most likely be deprecated and removed at some point.

### DIFF
--- a/src/frequenz/client/reporting/_client.py
+++ b/src/frequenz/client/reporting/_client.py
@@ -4,25 +4,22 @@
 """Client for requests to the Reporting API."""
 
 from collections import namedtuple
+from collections.abc import AsyncIterator, Iterator
 from dataclasses import dataclass
 from datetime import datetime
-from typing import Any, AsyncIterator, Awaitable, Iterator, Type, cast
+from typing import Any, cast
 
 import grpc.aio as grpcaio
 
 # pylint: disable=no-name-in-module
-from frequenz.api.common.v1.metrics.metric_sample_pb2 import Metric as PBMetric
 from frequenz.api.common.v1.microgrid.microgrid_pb2 import (
     MicrogridComponentIDs as PBMicrogridComponentIDs,
 )
-from frequenz.api.common.v1.pagination.pagination_params_pb2 import (
-    PaginationParams as PBPaginationParams,
+from frequenz.api.reporting.v1.reporting_pb2 import (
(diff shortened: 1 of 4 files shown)


src/frequenz/client/reporting/_client.py
```diff
@@ -248,71 +224,33 @@ def dt2ts(dt: datetime) -> PBTimestamp:
...
             )
-            if not response or response.is_empty():
-                break
-
-            yield response
 
-            page_token = response.next_page_token
-            if not page_token:
-                break
+            async for response in stream:
```
cwasicki, line 250: This `async` causes a mypy failure: https://github.com/frequenz-floss/frequenz-client-reporting-python/actions/runs/9495771269/job/26168719378?pr=67#step:8:22

llucax: grpcio has wrong type hints, it is probably that, look at:

https://github.com/frequenz-floss/frequenz-client-microgrid-python/blob/79d3f11cc6c836aef76203c896a912558bfe51a3/src/frequenz/client/microgrid/_client.py#L103-L109

for example.

cwasicki: The cast doesn't seem to solve the problem. I have added a type ignore to this line.

cwasicki:
Updated after @llucax suggestion to cast to `AsyncIterator` instead.

src/frequenz/client/reporting/_client.py
```diff
@@ -6,23 +6,19 @@
 from collections import namedtuple
 from dataclasses import dataclass
 from datetime import datetime
-from typing import Any, AsyncIterator, Awaitable, Iterator, Type, cast
+from typing import Any, AsyncIterator, Iterator, Type, cast
```
llucax: Nitpick, but those classes were added to the `typing` module when the type hints system wasn't mature enough and regular classes couldn't be used for typing I think. They are deprecated in typing since Python 3.9. For `Type` you can use the built-in `type`.

```suggestion
from typing import Any, cast
from collections.abc import AsyncIterator, Iterator
```

image

cwasicki: Fixed

flora-hofmann-frequenz (approved):

PR #69, opened 2024-06-13
Install example command line tool as reporting-cli

cwasicki:
* Install command line tool as `reporting-cli`: The command line tool from the example is moved to the package and can be called via `reporting-cli` or `python -m frequenz.client.reporting` from the command line.
* The command-line argument `--display` is renamed to `--format` for `reporting-cli`.
* The `reporting-cli` tool can output CSV format, dataframe output is removed.


RELEASE_NOTES.md
```diff
@@ -6,6 +6,12 @@
 
 ## Upgrading
 
+* Install command line tool as `reporting-cli`: The command line tool from the exampls is moved
```
Marenz: exampls typo

cwasicki: fixed

Marenz (approved):

Marenz:
Just to mention it: @llucax was less thrilled about having the client in __main__, but for now I think it's fine.

cwasicki:
> Just to mention it: @llucax was less thrilled about having the client in main, but for now I think it's fine.

Discussed with @llucax that we continue with this for now and follow-up on the structure subsequently.

PR #71, opened 2024-06-13
Update README with reporting-cli usage instructions

cwasicki:


README.md
```diff
@@ -30,7 +30,7 @@ Please also refer to [examples](https://github.com/frequenz-floss/frequenz-clien
 
 ```bash
 # Choose the version to install
-VERSION=0.4.0
```
Marenz, line 33: If you always wanna have the latest version here, does it make sense to specify any at all?

Marenz (approved):

README.md
cwasicki: Do you mean this should be automated? I think I would still specify it explicitly to keep the instructions in sync with the version.

llucax: It could be done via macros, but even if you don't want to do that, I think you can just say "to install the latest version: .... If you want to install an older version user: ....".

cwasicki: Isn't that obvious from the comment?

llucax: I meant so you can remove the version to avoid having to update it on every release.

cwasicki: Ah, I would do that later when the client is more stable. At the moment I find it important to be explicit which version is required for the examples in the README. Of course we ideally should update it with each release, but this is robust if they go out of sync.

PR #75, opened 2024-06-19
feat: Add secure connection using gRPC SSL credentials

sandovalrr:


Marenz:
Short discussion summary with Richard:

* Possibly easier to pass in the channel from outside to init
* Alternatively a flag "ssl" or "tls" that can be given in init?

It's certainly desirable to be able to test insecure channels, but the tests/ module could also just mock those parts.

llucax:
The new (unreleased) version of https://github.com/frequenz-floss/frequenz-client-base-python/ should help with this.

Maybe you can try out the branch and if it works well we can release it.

flora-hofmann-frequenz:
> Maybe you can try out the branch and if it works well we can release it.

We will look into this, but currently blocked and thus updated this PR first.

flora-hofmann-frequenz:
Closed as we continue working on #76

PR #89, opened 2024-07-26
Integrate BaseClient into reporting client

flora-hofmann-frequenz:
Still need to work on `__init__ method from base class 'BaseApiClient' is not called (super-init-not-called)`

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -33,6 +33,7 @@ dependencies = [
   "grpcio >= 1.54.2, < 2",
   "grpcio-tools >= 1.54.2, < 2",
   "protobuf >= 4.25.3, < 6",
+  "frequenz-client-base[grpcio] >= 0.5.0, < 0.6.0",
 ]
 dynamic = ["version"]
 
--- a/src/frequenz/client/reporting/_client.py
+++ b/src/frequenz/client/reporting/_client.py
@@ -7,9 +7,8 @@ from collections import namedtuple
 from collections.abc import AsyncIterator, Iterator
 from dataclasses import dataclass
 from datetime import datetime
-from typing import Any, cast
+from typing import cast
 
-import grpc
 import grpc.aio as grpcaio
 
 # pylint: disable=no-name-in-module
@@ -27,10 +26,9 @@ from frequenz.api.reporting.v1.reporting_pb2 import (
 )
 from frequenz.api.reporting.v1.reporting_pb2 import TimeFilter as PBTimeFilter
 from frequenz.api.reporting.v1.reporting_pb2_grpc import ReportingStub
+from frequenz.client.base.client import BaseApiClient
 from frequenz.client.common.metric import Metric
 from google.protobuf.timestamp_pb2 import Timestamp as PBTimestamp
-
-# pylint: enable=no-name-in-module
 
 MetricSample = namedtuple(
     "MetricSample", ["timestamp", "microgrid_id", "component_id", "metric", "value"]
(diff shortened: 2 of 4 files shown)


src/frequenz/client/reporting/_client.py
```diff
@@ -244,48 +245,15 @@ def dt2ts(dt: datetime) -> PBTimestamp:
...
                     request, metadata=self._metadata
                 ),
             )
-            # grpc.aio is missing types and mypy thinks this is not
-            # async iterable, but it is.
             async for response in stream:
                 if not response:
                     break
-
                 yield ComponentsDataBatch(response)
```
llucax, line 250: At some point we might want to consider to provide a way to wrap these streaming methods too in `client-base`, so we don't need to have these casts everywhere, which are also error prone.

flora-hofmann-frequenz:
I was trying to initialise with `super` from the `BaseClient`, but get an error as the service_address does not have the `grpc://` format that the server_url requests. Do you have any advice on this @llucax?

llucax:
Can you provide a more complete example? From what you say it sounds like you are passing an invalid `server_url`. The `server_url` should follow this format: `grpc://hostname[:port][?ssl=false]`, as described in [`parse_grpc_uri() `](https://frequenz-floss.github.io/frequenz-client-base-python/v0.5/reference/frequenz/client/base/channel/#frequenz.client.base.channel.parse_grpc_uri).

I see we don't have that defined in the [`ApiClientBase.__init__()`](https://frequenz-floss.github.io/frequenz-client-base-python/v0.5/reference/frequenz/client/base/client/#frequenz.client.base.client.BaseApiClient.__init__) docs, maybe we should document it there too.

flora-hofmann-frequenz:
@llucax: I haven't worked on the tests yet. I was wondering if you could give me feedback first and if we should change `SSL` to True for the connection first?

src/frequenz/client/reporting/_client.py
```diff
@@ -98,7 +98,7 @@ def __iter__(self) -> Iterator[MetricSample]:
                 )
 
 
-class ReportingApiClient:
+class ReportingApiClient(BaseApiClient[ReportingStub, grpcaio.Channel]):
     """A client for the Reporting service."""
 
     def __init__(self, service_address: str, key: str | None = None) -> None:
```
llucax: You should accept a URL here instead too, and just blindly pass the `server_url` to the `super().__init__()`. The end user needs to decide if they want to use SSL or not.

src/frequenz/client/reporting/_client.py
```diff
@@ -27,10 +26,11 @@
 )
 from frequenz.api.reporting.v1.reporting_pb2 import TimeFilter as PBTimeFilter
 from frequenz.api.reporting.v1.reporting_pb2_grpc import ReportingStub
+from frequenz.client.base.client import BaseApiClient
 from frequenz.client.common.metric import Metric
 from google.protobuf.timestamp_pb2 import Timestamp as PBTimestamp
 
-# pylint: enable=no-name-in-module
+StubT = TypeVar("StubT")
```
llucax: I don't know why this was added but it's not needed.

```suggestion
```

flora-hofmann-frequenz:
Will be updated with changes from [PR](https://github.com/frequenz-floss/frequenz-client-base-python/pull/67).

llucax:
> Will be updated with changes from [PR](https://github.com/frequenz-floss/frequenz-client-base-python/pull/67).

I suggest doing this in a future PR when a new version is released. We don't need to block this PR until that happens.

tests/test_client_reporting.py
```diff
@@ -2,27 +2,25 @@
...
-async def test_client_initialization(mock_channel: MagicMock) -> None:
-    """Test that the client initializes the channel."""
-    client = ReportingApiClient("localhost:50051")  # noqa: F841
-    mock_channel.assert_called_once_with("localhost:50051", ANY)
+async def test_client_initialization() -> None:
+    """Test that the client initializes the BaseApiClient with grpcaio.Channel."""
+    with patch.object(BaseApiClient, "__init__", return_value=None) as mock_base_init:
+        client = ReportingApiClient("http://localhost:50051")  # noqa: F841
+        mock_base_init.assert_called_once_with(
+            "http://localhost:50051", ReportingStub, grpcaio.Channel
```
llucax: This should be `gprc://localhost:50051` otherwise it will fail.

llucax: Right, it is not failing because it is mocked :P

Not sure what this test is testing, only that `super().__init__()` is called?

pyproject.toml
```diff
@@ -33,6 +33,7 @@ dependencies = [
   "grpcio >= 1.54.2, < 2",
   "grpcio-tools >= 1.54.2, < 2",
   "protobuf >= 4.25.3, < 6",
+  "frequenz-client-base >=0.5",
```
Marenz: The [Release Notes](https://github.com/frequenz-floss/frequenz-client-base-python/releases/tag/v0.4.0) mention that one should use 
```suggestion
  "frequenz-client-base[grpcio] >= 0.5.0, < 0.6.0",
```

tests/test_client_reporting.py
flora-hofmann-frequenz: Yes, we had the same test before and I just changed it to use `super().__init__()`.

Marenz (approved):

PR #93, opened 2024-09-03
Add helper function to calculate energy metric

flora-hofmann-frequenz:
We have a client request to read out energy metrics. Since the inverter / meter readings are rather unreliable, we want to add a helper function to calculate the `energy metric`.

### DIFF
--- a/src/frequenz/client/reporting/_client.py
+++ b/src/frequenz/client/reporting/_client.py
@@ -40,6 +40,11 @@ simplicity of a tuple. This data type can be easily used to create a numpy array
 or a pandas DataFrame.
 """
 
+CumulativeEnergy = namedtuple(
+    "CumulativeEnergy", ["start_time", "end_time", "consumption", "production"]
+)
+"""Type for cumulative energy consumption and production over a specified time."""
+
 
 @dataclass(frozen=True)
 class ComponentsDataBatch:
@@ -252,3 +257,82 @@ class ReportingApiClient(BaseApiClient[ReportingStub, grpcaio.Channel]):
         except grpcaio.AioRpcError as e:
             print(f"RPC failed: {e}")
             return
+
+
+# pylint: disable=too-many-arguments
+async def cumulative_energy(
+    client: ReportingApiClient,
+    microgrid_id: int,
+    component_id: int,
+    start_time: datetime,
+    end_time: datetime,
+    use_active_power: bool,
+    resolution: int | None = None,
+) -> CumulativeEnergy:
+    """
+    Calculate the cumulative energy consumption and production over a specified time range.
+
+    Args:
+        client: The client used to fetch the metric samples from the Reporting API.
+        microgrid_id: The ID of the microgrid.
+        component_id: The ID of the component within the microgrid.
+        start_time: The start date and time for the period.
+        end_time: The end date and time for the period.
+        use_active_power: If True, use the 'AC_ACTIVE_POWER' metric.
+                          If False, use the 'AC_ACTIVE_ENERGY' metric.
+        resolution: The resampling resolution for the data, represented in seconds.
+                    If None, no resampling is applied.
+
+    Returns:
+        EnergyMetric: A named tuple with start_time, end_time, consumption, and production.
+    """
+    metric = Metric.AC_ACTIVE_POWER if use_active_power else Metric.AC_ACTIVE_ENERGY
+
+    metric_samples = [
+        sample
+        async for sample in client.list_microgrid_components_data(
+            microgrid_components=[(microgrid_id, [component_id])],
+            metrics=metric,
+            start_dt=start_time,
+            end_dt=end_time,
+            resolution=resolution,
+        )
+    ]
+
+    values = [
+        sample.value
+        for sample in metric_samples
+        if start_time <= sample.timestamp <= end_time
+    ]
+
+    if values:
+        if use_active_power:
+            # Convert power to energy if using AC_ACTIVE_POWER
+            time_interval_hours = (
+                resolution or (end_time - start_time).total_seconds()
+            ) / 3600.0
+            consumption = sum(
+                max(e2 - e1, 0) * time_interval_hours
+                for e1, e2 in zip(values, values[1:])
+            )
+            production = sum(
+                min(e2 - e1, 0) * time_interval_hours
+                for e1, e2 in zip(values, values[1:])
+            )
+        else:
+            # Directly use energy values if using AC_ACTIVE_ENERGY
+            consumption = sum(
+                e2 - e1 for e1, e2 in zip(values, values[1:]) if e2 - e1 > 0
+            )
+            production = sum(
+                e2 - e1 for e1, e2 in zip(values, values[1:]) if e2 - e1 < 0
+            )
+    else:
+        consumption = production = 0.0
+
+    return CumulativeEnergy(
+        start_time=start_time,
+        end_time=end_time,
+        consumption=consumption,
+        production=production,
+    )
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -10,7 +10,7 @@
 
 ## New Features
 
-* Integration of the base client v0.5
+* Helper function to calculate energy metrics
 
 ## Bug Fixes
 


flora-hofmann-frequenz:
@llucax: This is a very first idea on how to include a helper function to the base client in reporting. Would you please add your input.

pyproject.toml
```diff
@@ -34,6 +34,7 @@ dependencies = [
   "grpcio-tools >= 1.54.2, < 2",
   "protobuf >= 4.25.3, < 6",
   "frequenz-client-base[grpcio] >= 0.5.0, < 0.6.0",
+  "pandas == 2.2.2"
```
llucax: Do we really want to make an API client depend on pandas? For me this sounds like it should be put in a higher level.

flora-hofmann-frequenz: No, not at all. It was just a first shot as I am unsure of the data structure in which we want to supply values. I am happy to move away from pandas. The best option may also be connected to data structure that we will use when exposing the states @cwasicki?

llucax (commented):
OK, so in terms of code structure, I would make the helper function a free function instead of a class method. And probably would take an `Iterator` or `Sequence` instead of a `list` if that's an option.

nkrtzr:
@flora-hofmann-frequenz just for my understanding, what happens if there is a gap in data for the power?

nkrtzr:
I also think that the customer would rather expect the total amount of energy that was consumed or delivered as output, rather than the energy for every timestamp. Something like a "Zählerstand" or at least the difference between the "Zählerstände"

flora-hofmann-frequenz:
> I also think that the customer would rather expect the total amount of energy that was consumed or delivered as output, rather than the energy for every timestamp. Something like a "Zählerstand" or at least the difference between the "Zählerstände"

@noah-kreutzer-frequenz can you please make an example for this? Isn't the `total amount` dependent on a certain timeframe?

nkrtzr:
> > I also think that the customer would rather expect the total amount of energy that was consumed or delivered as output, rather than the energy for every timestamp. Something like a "Zählerstand" or at least the difference between the "Zählerstände"
> 
> @noah-kreutzer-frequenz can you please make an example for this? Isn't the `total amount` dependent on a certain timeframe?

So rather than a Timeseries the looking like this

time | energy | power related to the time frame
-------- | -------- | --------
0:00  | 0.0   | 0
1:00   | 1.0  |  1
2:00   | 2.0  | 2
3:00   | 0.0  | 2
4:00   | 4.0  | 4
5:00   | 0.0  | 0

I think something like this is what the customer  would expect

time | total energy | power related to the time frame
-------- | -------- | --------
0:00  | 0.0   | 0
1:00   | 1.0  |  1
2:00   | 3.0  | 2
3:00   | 3.0  | 0
4:00   | 7.0  | 4
5:00   | 7.0  | 0

or even only the the last value

nkrtzr:
and potentially the possoiblity of seperating between consumed energy and delivered energy, so they don't cancel out :))

pyproject.toml
cwasicki: Agree, we shouldn't use pandas if not urgently required.

src/frequenz/client/reporting/_client.py
```diff
@@ -252,3 +253,33 @@ def dt2ts(dt: datetime) -> PBTimestamp:
...
+
+        ac_active_power_series = pd.Series(
+            data=[sample.value for sample in ac_active_power_samples],
+            index=[sample.timestamp for sample in ac_active_power_samples],
+        )
+
+        # Convert the timedelta to hours
+        hours = time_interval.total_seconds() / 3600.0
+
+        energy_wh_series = ac_active_power_series * hours
```
cwasicki: I don't think this is what is wanted. Usually the energy metric is integrated over time, so the power would need to be converted to energy and then cumulatively added.

src/frequenz/client/reporting/_client.py
```diff
@@ -252,3 +253,33 @@ def dt2ts(dt: datetime) -> PBTimestamp:
         except grpcaio.AioRpcError as e:
             print(f"RPC failed: {e}")
             return
+
+    @staticmethod
+    def calculate_energy_metric(
+        metric_samples: list[MetricSample], time_interval: timedelta
+    ) -> pd.Series:
+        """
+        Calculate the energy metric with AC_ACTIVE_POWER values and time interval.
```
cwasicki: I don't remember whether this should have been from power or the actual energy metrics. For latter, the only thing that needs to be done is to remove the resets. That could be achieved by taking the diff, removing the negatives (reset) and cumulatively sum them up again.

src/frequenz/client/reporting/_client.py
```diff
@@ -252,3 +253,33 @@ def dt2ts(dt: datetime) -> PBTimestamp:
         except grpcaio.AioRpcError as e:
             print(f"RPC failed: {e}")
             return
+
+    @staticmethod
+    def calculate_energy_metric(
+        metric_samples: list[MetricSample], time_interval: timedelta
```
cwasicki: I think this should be a wrapper around ` list_microgrid_components_data` with the same arguments but the metric, which is given.

src/frequenz/client/reporting/_client.py
```diff
@@ -252,3 +253,33 @@ def dt2ts(dt: datetime) -> PBTimestamp:
         except grpcaio.AioRpcError as e:
             print(f"RPC failed: {e}")
             return
+
+    @staticmethod
```
cwasicki: Good that this is staticmethod, I see this as part of a tooling on top of the client and not the client itself. For now that would be overkill though.

src/frequenz/client/reporting/_client.py
```diff
@@ -252,3 +253,33 @@ def dt2ts(dt: datetime) -> PBTimestamp:
...
+    def calculate_energy_metric(
```
cwasicki: Needs a clearer name, maybe `cumulative_energy`, `total_energy, `integrated_power` or so, depending on which approach we go for.

flora-hofmann-frequenz:
@cwasicki & @noah-kreutzer-frequenz please take a look at the energy metric logic again. I was trying to implement the following:

```
consumption = sum(
    e2 - e1 for e1, e2 in zip(filtered_samples, filtered_samples[1:])
    if e2 - e1 > 0
)
```
```
production = sum(
    e2 - e1 for e1, e2 in zip(filtered_samples, filtered_samples[1:])
    if e2 - e1 < 0
)
```
Alongside the previous comments and the option to use power or energy metrics (in case energy is not available).

src/frequenz/client/reporting/_client.py
```diff
@@ -252,3 +257,82 @@ def dt2ts(dt: datetime) -> PBTimestamp:
...
+            start_dt=start_time,
+            end_dt=end_time,
+            resolution=resolution,
+        )
+    ]
+
+    values = [
+        sample.value
+        for sample in metric_samples
+        if start_time <= sample.timestamp <= end_time
```
cwasicki, line 305: Shouldn't happen anyway, but typically exclusive timestamps for the end time are used, i.e. `< end_time`.

src/frequenz/client/reporting/_client.py
```diff
@@ -252,3 +257,82 @@ def dt2ts(dt: datetime) -> PBTimestamp:
...
+    ]
+
+    if values:
+        if use_active_power:
+            # Convert power to energy if using AC_ACTIVE_POWER
+            time_interval_hours = (
+                resolution or (end_time - start_time).total_seconds()
```
cwasicki, line 312: The second part looks off to me since each power needs to be integrated over it's time period to energy, so would only work for given resolution. You could just use:
```
consumption = sum(
    m1.value * (m2.timestamp-m1.timestamp).total_seconds() 
    for m in zip(metric_samples, metric_samples[1:] 
    if m1.value > 0) / 3600.0
```

src/frequenz/client/reporting/_client.py
cwasicki: Why do we actually need to extract the values again and not just loop over the `metric_samples` (see example for power).

src/frequenz/client/reporting/_client.py
```diff
@@ -252,3 +257,82 @@ def dt2ts(dt: datetime) -> PBTimestamp:
         except grpcaio.AioRpcError as e:
             print(f"RPC failed: {e}")
             return
+
+
+# pylint: disable=too-many-arguments
+async def cumulative_energy(
```
cwasicki, line 263: Maybe this should be part of the client for now, otherwise it could be a separate module.

src/frequenz/client/reporting/_client.py
```diff
@@ -252,3 +257,82 @@ def dt2ts(dt: datetime) -> PBTimestamp:
...
+    client: ReportingApiClient,
```
cwasicki, line 264: If it's not part of the client, it needs an URL and key. I guess part of the client makes things simpler now. 

@llucax Any thoughts on this? This helper function IMO goes beyond a thin client and could be seen as additional tooling on top of the client. We do think it is overkill to start that now though and wait for further cases.

llucax: > If it's not part of the client, it needs an URL and key. I guess part of the client makes things simpler now.

No, it just needs a client, which you should have already instantiated. It really exactly the same, is just a syntax issue. The method would be `async def cumulative_energy(self: ReportingApiClient,...)` the same, it is just called `self` by convention. You can actually also call a method like this `ReportingApiClient.cumulative_energy(client, ...)`.

I like utilities being separated because it keeps the client interface clear, otherwise is really hard to draw a line between what should be part of the client what shouldn't.

> @llucax Any thoughts on this? This helper function IMO goes beyond a thin client and could be seen as additional tooling on top of the client. We do think it is overkill to start that now though and wait for further cases.

I think we should adopt the https://github.com/frequenz-floss/frequenz-dispatch-python approach if there is a room for a higher level interface. At least IMHO that worked out well. In *dispatch* we basically have the https://github.com/frequenz-floss/frequenz-client-dispatch-python as a thin wrapper over the API and then `frequenz-dispatch-python` provides a higher level interface, which makes a lot of assumptions about how the API should be used, so it is much less flexible, but much easier to use for 99% of the cases (scientifically proven :P).

flora-hofmann-frequenz: Okay, I think it will be best to set up `frequenz-reporting-python` and move the energy metric there :-)

flora-hofmann-frequenz:
This PR will be closed as it moved to [this PR](https://github.com/frequenz-floss/frequenz-reporting-python/pull/7) in a new repo.

PR #97, opened 2024-09-13
Add support for states in client and cli tool 

### LINKED ISSUE #25: Support request include option for states
### What's needed?

The include option for states should be exposed, state data should be converted from the response and integrated with the output data structure.

### Proposed solution

_No response_

### Use cases

_No response_

### Alternatives and workarounds

_No response_

### Additional context

For output data type see also requirements from: https://github.com/frequenz-floss/frequenz-client-reporting-python/issues/20.

cwasicki:
States can now be requested via the client and are provided through the flat iterator. They can be identified via their category `state`, `warning` and `error`, respectively. Each individual state is provided as its own sample.

Support for states is also added to the CLI tool via the `--states` flag. As of now requests for states without any metrics are not yet supported by the service.

Fixes https://github.com/frequenz-floss/frequenz-client-reporting-python/issues/25

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -10,7 +10,11 @@
 
 ## New Features
 
-<!-- Here goes the main new features and examples or instructions on how to use them -->
+* States can now be requested via the client and are provided through the flat iterator.
+  They can be identified via their category `state`, `warning` and `error`, respectively.
+  Each individual state is provided as its own sample.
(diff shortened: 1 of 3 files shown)


flora-hofmann-frequenz (approved):
LGTM - I think it makes sense to keep the states information in the client and not add it to the interface repo as it only exposes them.

PR #98, opened 2024-09-17
Add support for bounds in client

### LINKED ISSUE #20: Support request option for bounds
### What's needed?

The include option for bounds should be exposed, bound data should be converted from the response and integrated with the output data structure.

### Proposed solution

_No response_

### Use cases

_No response_

### Additional context

For output data type see also requirements from: https://github.com/frequenz-floss/frequenz-client-reporting-python/issues/25.

cwasicki:
Bounds can now be requested via the client and are provided through the flat iterator. They can be identified via their category `metric_bound[i]_{upper,lower}`. Each individual bound is provided as its own sample.

Support for states and bound is also added to the CLI tool via the `--bounds` flag.

Fixes https://github.com/frequenz-floss/frequenz-client-reporting-python/issues/20.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -13,8 +13,11 @@
 * States can now be requested via the client and are provided through the flat iterator.
   They can be identified via their category `state`, `warning` and `error`, respectively.
   Each individual state is provided as its own sample.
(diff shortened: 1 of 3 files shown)


flora-hofmann-frequenz (approved):
LGTM

Issue #79, opened 2024-06-21
Rename resolution to resampling_period


flora-hofmann-frequenz:
### What's needed?

Change resolution to sampling_period for easier understanding what the resolution parameter does.

### Proposed solution

resolution: int to sampling_period: timedelta

flora-hofmann-frequenz:
Should we still make this change @cwasicki? If so I could tackle it next week.

cwasicki:
Yes, to me this makes sense.

flora-hofmann-frequenz:
In my first shot at this, I assume that we only want to change the client to be able to handle `timedelta`, not `PBResamplingOptions(resolution=...)`

PR #105, opened 2024-10-30
Change resolution to resample_period

flora-hofmann-frequenz:
Addressing Issue #79.

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,8 +50,9 @@ client = ReportingApiClient(server_url=SERVER_URL, key=API_KEY)
 ```
 
 Besides the microgrid_id, component_ids, and metrics, start, and end time,
-you can also set the sampling period for resampling using the `resolution` parameter.
-For example, to resample data every 15 minutes, use a `resolution` of 900 seconds. The default is 1 second.
+you can also set the sampling period for resampling using the `resampling_period`
+parameter. For example, to resample data every 15 minutes, use a `resampling_period`
+of timedelta(minutes=15).
 
 ### Query metrics for a single microgrid and component:
 
@@ -64,7 +65,7 @@ data = [
         metrics=[Metric.AC_ACTIVE_POWER, Metric.AC_REACTIVE_POWER],
         start_dt=datetime.fromisoformat("2024-05-01T00:00:00"),
         end_dt=datetime.fromisoformat("2024-05-02T00:00:00"),
-        resolution=1,
+        resampling_period=timedelta(seconds=1),
     )
 ]
 ```
@@ -91,7 +92,7 @@ data = [
         metrics=[Metric.AC_ACTIVE_POWER, Metric.AC_REACTIVE_POWER],
         start_dt=datetime.fromisoformat("2024-05-01T00:00:00"),
         end_dt=datetime.fromisoformat("2024-05-02T00:00:00"),
-        resolution=1,
+        resampling_period=timedelta(seconds=1),
         states=False, # Set to True to include state data
         bounds=False, # Set to True to include metric bounds data
     )
--- a/src/frequenz/client/reporting/__main__.py
+++ b/src/frequenz/client/reporting/__main__.py
@@ -5,7 +5,7 @@
 
 import argparse
 import asyncio
-from datetime import datetime
+from datetime import datetime, timedelta
 from pprint import pprint
 from typing import AsyncIterator
 
@@ -57,7 +57,12 @@ def main() -> None:
         help="End datetime in YYYY-MM-DDTHH:MM:SS format",
         required=True,
     )
-    parser.add_argument("--resolution", type=int, help="Resolution", default=None)
+    parser.add_argument(
+        "--resampling_period_s",
+        type=int,
+        help="Resampling period in seconds (integer, rounded to avoid subsecond precision issues).",
+        default=None,
+    )
     parser.add_argument("--psize", type=int, help="Page size", default=1000)
     parser.add_argument(
         "--format", choices=["iter", "csv", "dict"], help="Output format", default="csv"
@@ -76,7 +81,7 @@ def main() -> None:
             metric_names=args.metrics,
             start_dt=args.start,
             end_dt=args.end,
-            resolution=args.resolution,
+            resampling_period_s=args.resampling_period_s,
             states=args.states,
             bounds=args.bounds,
             service_address=args.url,
@@ -94,7 +99,7 @@ async def run(
     metric_names: list[str],
     start_dt: datetime,
     end_dt: datetime,
-    resolution: int,
+    resampling_period_s: int | None,
     states: bool,
     bounds: bool,
     service_address: str,
@@ -109,7 +114,7 @@ async def run(
         metric_names: list of metric names
         start_dt: start datetime
         end_dt: end datetime
-        resolution: resampling resolution in sec
+        resampling_period_s: The period for resampling the data.
         states: include states in the output
         bounds: include bounds in the output
         service_address: service address
@@ -131,13 +136,19 @@ async def run(
         Returns:
             Iterator over single metric samples
         """
+        resampling_period = (
+            timedelta(seconds=resampling_period_s)
+            if resampling_period_s is not None
+            else None
+        )
+
         return client.list_single_component_data(
             microgrid_id=microgrid_id,
             component_id=component_id,
             metrics=metrics,
             start_dt=start_dt,
             end_dt=end_dt,
-            resolution=resolution,
+            resampling_period=resampling_period,
             include_states=states,
             include_bounds=bounds,
         )
--- a/src/frequenz/client/reporting/_client.py
+++ b/src/frequenz/client/reporting/_client.py
@@ -6,7 +6,7 @@
 from collections import namedtuple
 from collections.abc import AsyncIterator, Iterable, Iterator
 from dataclasses import dataclass
-from datetime import datetime
+from datetime import datetime, timedelta
 from typing import cast
 
 import grpc.aio as grpcaio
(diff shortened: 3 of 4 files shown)


Marenz (dismissed):

src/frequenz/client/reporting/_client.py
```diff
@@ -160,7 +160,7 @@ async def list_single_component_data(
         metrics: Metric | list[Metric],
         start_dt: datetime,
         end_dt: datetime,
-        resolution: int | None,
+        resampling_period: timedelta,
```
cwasicki: This can also be `None` if no resampling is desired.

README.md
```diff
@@ -50,8 +50,9 @@ client = ReportingApiClient(server_url=SERVER_URL, key=API_KEY)
 ```
 
 Besides the microgrid_id, component_ids, and metrics, start, and end time,
-you can also set the sampling period for resampling using the `resolution` parameter.
-For example, to resample data every 15 minutes, use a `resolution` of 900 seconds. The default is 1 second.
+you can also set the sampling period for resampling using the `resampling_period`
```
cwasicki, line 53: Just noting that the SDK uses `sampling_period`, but I think the term you chose is more accurate, so I would keep it.

src/frequenz/client/reporting/_client.py
flora-hofmann-frequenz: I get a mypy error when I set it to `None` and the use `total_seconds()` in the streaming filter. Not quite sure how to work around it?

flora-hofmann-frequenz: > _client.py:303: error: Item "None" of "timedelta | None" has no attribute "total_seconds"  [union-attr]

cwasicki: If it's None you can pass that to the API, otherwise you convert it. E.g.
```
resolution = td.total_seconds() if sp is not None else None
```

flora-hofmann-frequenz: I have added this in.

src/frequenz/client/reporting/_client.py
```diff
@@ -299,7 +299,13 @@ def dt2ts(dt: datetime) -> PBTimestamp:
 
         stream_filter = PBReceiveMicrogridComponentsDataStreamRequest.StreamFilter(
             time_filter=time_filter,
-            resampling_options=PBResamplingOptions(resolution=resolution),
+            resampling_options=PBResamplingOptions(
+                resolution=(
+                    int(resampling_period.total_seconds())
```
Marenz: I am wondering, do we want `int()` which cuts off the fractional part or rather `round()` which .. well.. rounds ;)

src/frequenz/client/reporting/__main__.py
```diff
@@ -56,7 +56,9 @@ def main() -> None:
         help="End datetime in YYYY-MM-DDTHH:MM:SS format",
         required=True,
     )
-    parser.add_argument("--resolution", type=int, help="Resolution", default=None)
+    parser.add_argument(
+        "--resampling_period", type=int, help="Resampling period", default=None
```
Marenz: Is anywhere explained what unit this parameter expects? And is it desired that sub-second durations (if those are seconds) are not supported (int)?

src/frequenz/client/reporting/_client.py
flora-hofmann-frequenz: I am not sure, but we had resolution : int beforehand, that's why I kept it this way

flora-hofmann-frequenz: But it also makes sense, cause the client/user would likely use 1min, 15min and probably nothing too complicated in between. If it needs to be as it is read out then one can just leave it defaulting to None.

src/frequenz/client/reporting/__main__.py
flora-hofmann-frequenz: This is the reference to the proto file: https://github.com/frequenz-floss/frequenz-api-reporting/blob/72510043a64517dd0f7eba1cd6ea8b0fa6f898ae/proto/frequenz/api/reporting/v1/reporting.proto#L81

src/frequenz/client/reporting/_client.py
Marenz: sure, but due to floating point, you could end up with an number like 4.999999 in there and instead of 5 you'll get 4 this way.

src/frequenz/client/reporting/__main__.py
Marenz: Right, but when calling this program with "--help", I won't have the proto buf in mind and how something maps exactly from that to my command line parameter?

src/frequenz/client/reporting/_client.py
flora-hofmann-frequenz: But if the proto defines `uint32` should we then not also use `int` here?

src/frequenz/client/reporting/__main__.py
flora-hofmann-frequenz: ahhh sorry, you mean i should add something like: `help="Resampling period in seconds (integer, no sub-second durations)."` (obviously pending our discussion above).

src/frequenz/client/reporting/_client.py
cwasicki: You would still use int because the API doesn't support sub-second resolution. But int just cuts off the fractional part, so what @Marenz describes is a problem with floating points in general where you cannot always represent integers exactly. So I agree with rounding here.

cwasicki: >  where you cannot always represent integers exactly.

In practice this should less of a concern in the range we use this parameter, maybe more if the value is a result of floating point operations I guess. But it's still good practice :)

src/frequenz/client/reporting/__main__.py
Marenz: The parameter here is int, so that "sub second" part is not needed. But you convert it later to a time-delta which then converts it to seconds, but in floating point. And that last conversion is only the part where the problem can happen. Though rather theoretical maybe, depending what the source of the value is :)

src/frequenz/client/reporting/_client.py
```diff
@@ -172,7 +172,7 @@ async def list_single_component_data(
             metrics: The metric name or list of metric names.
             start_dt: The start date and time.
             end_dt: The end date and time.
-            resolution: The resampling resolution for the data, represented in seconds.
+            resampling_period: The period for resampling the data, specified as `timedelta` object.
```
Marenz: No need to change it now, but for future, I think the extra type info isn't required in the comment, the type hints already give us that :)

Marenz (dismissed):

src/frequenz/client/reporting/_client.py
flora-hofmann-frequenz: Good point! I have taken it out since I had to rebase on v0.x.x to get rid of the merge commit in any case.

Marenz (dismissed):
:fireworks:

src/frequenz/client/reporting/__main__.py
```diff
@@ -93,7 +98,7 @@ async def run(
     metric_names: list[str],
     start_dt: datetime,
     end_dt: datetime,
-    resolution: int,
+    resampling_period: timedelta,
```
cwasicki: You are passing an int (and in fact also `None` is possible, which is missing here). I would leave this and convert to timedelta (if set) below.

flora-hofmann-frequenz: Sorry, I am not entirely sure that I understand this comment. I tried the following change:

>     resampling_period: int | None,
>     states: bool,
>     bounds: bool,
>     service_address: str,
>     key: str,
>     fmt: str,
> ) -> None:
>     """Test the ReportingApiClient.
> 
>     Args:
>         microgrid_id: microgrid ID
>         component_id: component ID
>         metric_names: list of metric names
>         start_dt: start datetime
>         end_dt: end datetime
>         resampling_period: The period for resampling the data.
>         states: include states in the output
>         bounds: include bounds in the output
>         service_address: service address
>         key: API key
>         fmt: output format
> 
>     Raises:
>         ValueError: if output format is invalid
>     """
>     client = ReportingApiClient(service_address, key)
> 
>     metrics = [Metric[mn] for mn in metric_names]
> 
>     if resampling_period is not None:
>         resampling_period = timedelta(seconds=resampling_period)
> 
>     def data_iter() -> AsyncIterator[MetricSample]:
>         """Iterate over single metric.
> 
>         Just a wrapper around the client method for readability.
> 
>         Returns:
>             Iterator over single metric samples
>         """
>         return client.list_single_component_data(
>             microgrid_id=microgrid_id,
>             component_id=component_id,
>             metrics=metrics,
>             start_dt=start_dt,
>             end_dt=end_dt,
>             resampling_period=resampling_period,
>             include_states=states,
>             include_bounds=bounds,
>         )

But now I am getting (expected) mypy errors?

> src/frequenz/client/reporting/__main__.py:131: error: Incompatible types in assignment (expression has type "timedelta", variable has type "int | None")  [assignment]
> src/frequenz/client/reporting/__main__.py:147: error: Argument "resampling_period" to "list_single_component_data" of "ReportingApiClient" has incompatible type "int | None"; expected "timedelta | None"  [arg-type]

Marenz: Here in the `run` method, the type should be `int | None`, not in the `ReportingApiClient` Constructor if I read it correctly.
You then only need to convert it to timedeltta in line 144. 

I would also recommend renaming the parameter to `resampling_period_s` to indicate that it is seconds in the name right away. Once it is a timedelta that extra `_s` won't be needed anymore

cwasicki: >Here in the run method, the type should be int | None

Yes, and below:

```
resampling_period = timedelta(seconds=resampling_period_s) if resampling_period_s is not None or None
return client.list_single_component_data(
            microgrid_id=microgrid_id,
            component_id=component_id,
            metrics=metrics,
            start_dt=start_dt,
            end_dt=end_dt,
            resampling_period=resampling_period,
            include_states=states,
            include_bounds=bounds,
        )
```

cwasicki (approved):

PR #109, opened 2024-11-12
Support requesting no metrics in CLI tools

cwasicki:
The service will soon support requesting component states without a metric. This adds support for this feature in the CLI tool.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -10,6 +10,7 @@
 
 ## New Features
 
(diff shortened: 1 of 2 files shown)


cwasicki:
FYI @cyiallou

cyiallou:
Looks good to me but apparently the commit needs to be signed

cwasicki:
Updated

cyiallou (approved):

PR #114, opened 2024-11-27
Upgrade to reporting API v0.5

flora-hofmann-frequenz:
This update includes `MetricConnections` in the request message, that comes with a breaking change in the API specs. To ensure compatibility, both the client and service must be updated simultaneously. Specifically:
* This and later versions of the client can only be used with the latest reporting API service >=v0.7 but not with older versions of the service.
* Older versions of the client can only be used with the reporting API service versions < v0.7 but no later versions.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -6,7 +6,10 @@
 
 ## Upgrading
 
-<!-- Here goes notes on how to upgrade from previous versions, including deprecations and what they should be replaced with -->
+This update comes with a breaking change, that originates from a breaking change in the API specs. Specifically:
+* This and later versions of the client can only be used with the latest reporting API service >=v0.7 but not with older versions of the service.
+* Older versions of the client can only be used with the reporting API service versions < v0.7 but no later versions.
+To ensure compatibility, both the client and service must be updated simultaneously.
 
 ## New Features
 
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -28,7 +28,7 @@ requires-python = ">= 3.11, < 4"
 dependencies = [
   "typing_extensions >= 4.6.1, < 5",
   "grpc-stubs == 1.53.0.5",
-  "frequenz-api-reporting >= 0.2.0, < 0.5",
+  "frequenz-api-reporting >= 0.2.0, < 0.6",
```
cwasicki: The lower version needs to be updated where `MetricConnections` was introduced, otherwise this fails.

src/frequenz/client/reporting/_client.py
```diff
@@ -123,11 +128,8 @@ def __iter__(self) -> Iterator[MetricSample]:
                     "warning": state.warnings,
                     "error": state.errors,
                 }.items():
-                    # Skip if the category is not present
```
cwasicki: Why are these comments removed?

RELEASE_NOTES.md
```diff
@@ -6,7 +6,7 @@
 
 ## Upgrading
 
-<!-- Here goes notes on how to upgrade from previous versions, including deprecations and what they should be replaced with -->
+* Update to use frequenz-api-reporting 0.6
```
cwasicki: This needs more explanation because this has a dependency with the service.

cwasicki (approved):

Issue #23, opened 2024-03-21
Support real-time streaming requests


cwasicki:
### What's needed?

The client should support streaming requests in addition to the existing support for historical data.

flora-hofmann-frequenz:
Streaming endpoint is used for historical data from [Release v0.6.0](https://github.com/frequenz-floss/frequenz-client-reporting-python/releases/tag/v0.6.0)

cwasicki:
Fixed by #116.

PR #116, opened 2024-11-27
Support for streaming

cwasicki:
This adds support for streaming: By omitting the end date in the request, the client will return any historical data from timestamp until now and keep streaming new data as it arrives.

Also the start date can be omitted which let's the data start at the earliest time stamp that is available.


flora-hofmann-frequenz (approved):
LGTM!

cwasicki:
Fixes #23.

PR #134, opened 2025-01-21
Support requests for multiple components and formulas via CLI tool

cwasicki:
Support data requests for multiple component IDs and passing formula strings as a `cid` argument to the command line tool.


flora-hofmann-frequenz (approved):
Looks good to me. 
The formula would again be of type "#234 + #263" for example, right?

cwasicki:
> Looks good to me.
The formula would again be of type "#234 + #263" for example, right?

Yes, for v1.0 we need to update the information on this on the README.

PR #138, opened 2025-02-21
Ensure timezone-aware timestamps in formula endpoint result

cwasicki:
Required since the service responds with naive timestamps in UTC.


stefan-brus-frequenz (approved):
LGTM

PR #148, opened 2025-03-10
Update repo config w migration script

flora-hofmann-frequenz:


cwasicki (dismissed):
CI is failing but when solved this PR LGTM

llucax:
The error seems to be a regression in protobuf :disappointed:

It is supposed to be fixed in the next release: https://github.com/protocolbuffers/protobuf/issues/18096

It looks like it should be fixed in v28.3, I'm not sure who is requesting 28.2, but if you bump that to 28.3 I think it should work: https://github.com/protocolbuffers/protobuf/releases/tag/v28.3 ("Mute the minor version warning").

cwasicki (approved):

PR #154, opened 2025-03-27
Introduce GrpcStreamBroadcaster to enable keep_alive options

flora-hofmann-frequenz:
After a sensitivity analysis, the default `ChannelOptions()` of the base client are used (with the option to change them), as even with different configurations of `KeepAliveOptions()` of `interval` and `timeout`, the client sometimes has to retry.

### DIFF
--- a/src/frequenz/client/reporting/_client.py
+++ b/src/frequenz/client/reporting/_client.py
@@ -7,9 +7,7 @@ from collections import namedtuple
 from collections.abc import AsyncIterator, Iterable, Iterator
 from dataclasses import dataclass
 from datetime import datetime, timedelta, timezone
-from typing import cast
-
-import grpc.aio as grpcaio
+from typing import Any, AsyncIterable, cast
 
 # pylint: disable=no-name-in-module
 from frequenz.api.common.v1.microgrid.microgrid_pb2 import (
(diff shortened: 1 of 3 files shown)


src/frequenz/client/reporting/_client.py
```diff
@@ -367,21 +379,33 @@ def dt2ts(dt: datetime) -> PBTimestamp:
...
+            stream_name="microgrid-components-data-stream",
+            stream_method=stream_method,
+            transform=transform_response,
+            retry_strategy=None,
+        )
 
-        except grpcaio.AioRpcError as e:
-            print(f"RPC failed: {e}")
-            return
+        receiver = broadcaster.new_receiver()
```
cwasicki, line 406: Oh, we get a receiver here. We can have two methods, one returning an iterator as we have now, another returning a receiver. That would be useful for you, correct @phillip-wenig-frequenz ? 

Can be a subsequent PR though.

src/frequenz/client/reporting/_client.py
```diff
@@ -442,18 +465,26 @@ def dt2ts(dt: datetime) -> PBTimestamp:
...
+                yield response
+
+        broadcaster = GrpcStreamBroadcaster(
+            stream_name="aggregated-microgrid-data-stream",
+            stream_method=stream_method,
+            transform=transform_response,
+            retry_strategy=None,
+        )
+
+        receiver = broadcaster.new_receiver()
```
cwasicki, line 488: Same comment about the receiver method as above.

cwasicki (approved):
Thank you @flora-hofmann-frequenz !

src/frequenz/client/reporting/_client.py
flora-hofmann-frequenz: Will add it in a follow up PR! (Issue #155)

shsms: Actually you just need a function that returns a receiver, because the receiver is already an iterator.

Issue #155, opened 2025-03-28
Add receiver from the GRPCBroadcaster


flora-hofmann-frequenz:
### What's needed?

Add a receiver.

### Proposed solution

Not only iterator as we have now. See PR #154.

flora-hofmann-frequenz:
PR #154

PR #161, opened 2025-04-01
Add receiver method

flora-hofmann-frequenz:
Addresses Issue #155.  This implementation introduces an option (`stream_as_receiver`) to switch between using an AsyncIterator (default behavior) and a receiver. 

`stream_as_receiver=True` returns a receiver that can be shared across multiple consumers. Here is information about the underlying [GRPCBroadCaster definition](https://github.com/frequenz-floss/frequenz-client-base-python/blob/b72e84f258c4b39082bbe9e26933774cadb4d156/src/frequenz/client/base/streaming.py#L27C1-L138C42).

### DIFF
--- a/src/frequenz/client/reporting/_client.py
+++ b/src/frequenz/client/reporting/_client.py
@@ -222,6 +222,7 @@ class ReportingApiClient(BaseApiClient[ReportingStub]):
         resampling_period: timedelta | None,
         include_states: bool = False,
         include_bounds: bool = False,
+        stream_as_receiver: bool = False,
     ) -> AsyncIterator[MetricSample]:
         """Iterate over the data for a single metric.
 
@@ -234,6 +235,7 @@ class ReportingApiClient(BaseApiClient[ReportingStub]):
             resampling_period: The period for resampling the data.
             include_states: Whether to include the state data.
             include_bounds: Whether to include the bound data.
+            stream_as_receiver: Whether to return a receiver instead of using an iterator.
 
         Yields:
             A named tuple with the following fields:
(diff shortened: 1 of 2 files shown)


flora-hofmann-frequenz:
Please let me know if this is what you need in the application @cwasicki & @phillip-wenig-frequenz 

I'll still run a repo-config update, and then update the base branch here, but thought we could discuss already.

src/frequenz/client/reporting/_client.py
```diff
@@ -416,6 +432,7 @@ async def receive_aggregated_data(
         start: datetime | None,
         end: datetime | None,
         resampling_period: timedelta,
+        stream_as_receiver: bool = False,
```
cwasicki, line 435: I think the method should be split into two, one returning the receiver, and the other one calling the same but loops over the receiver and yields an iterator. 

Probably this method should return the receiver and we add a new method e.g. `iterate_aggregated_data` that yields the iterator. There is a big drawback thought, which is that this would be a breaking change. Maybe there are better options?

cwasicki:
As @shsms pointed out [here](https://github.com/frequenz-floss/frequenz-client-reporting-python/pull/154#discussion_r2026572915), we don't need two methods but just switch to return the receiver instead.

flora-hofmann-frequenz:
@cwasicki:  The function in PR #177 now just [returns the receiver](https://github.com/frequenz-floss/frequenz-client-reporting-python/blob/c593f14ca0c5b0ebce4df61b92ab2319a323fd98/src/frequenz/client/reporting/_client.py#L431). This means I can close this PR, or do I get this wrong?

cwasicki:
> @cwasicki: The function in PR https://github.com/frequenz-floss/frequenz-client-reporting-python/pull/177 now just [returns the receiver](https://github.com/frequenz-floss/frequenz-client-reporting-python/blob/c593f14ca0c5b0ebce4df61b92ab2319a323fd98/src/frequenz/client/reporting/_client.py#L431). This means I can close this PR, or do I get this wrong?

Exactly, this one can be closed.

PR #169, opened 2025-04-24
Add sensor ID

flora-hofmann-frequenz:
No aggregation function added yet.

### DIFF
--- a/src/frequenz/client/reporting/_client.py
+++ b/src/frequenz/client/reporting/_client.py
@@ -3,8 +3,8 @@
 
 """Client for requests to the Reporting API."""
 
-from collections import namedtuple
-from collections.abc import AsyncIterator, Iterable, Iterator
+from collections import abc, namedtuple
+from collections.abc import AsyncIterator, Iterator
 from dataclasses import dataclass
 from datetime import datetime, timedelta, timezone
 from typing import Any, AsyncIterable, cast
@@ -12,6 +12,9 @@ from typing import Any, AsyncIterable, cast
 # pylint: disable=no-name-in-module
 from frequenz.api.common.v1.microgrid.microgrid_pb2 import (
     MicrogridComponentIDs as PBMicrogridComponentIDs,
+)
+from frequenz.api.common.v1.microgrid.microgrid_pb2 import (
+    MicrogridSensorIDs as PBMicrogridSensorIDs,
 )
 from frequenz.api.reporting.v1.reporting_pb2 import (
     AggregationConfig as PBAggregationConfig,
@@ -31,6 +34,12 @@ from frequenz.api.reporting.v1.reporting_pb2 import (
 )
 from frequenz.api.reporting.v1.reporting_pb2 import (
     ReceiveMicrogridComponentsDataStreamResponse as PBReceiveMicrogridComponentsDataStreamResponse,
+)
+from frequenz.api.reporting.v1.reporting_pb2 import (
+    ReceiveMicrogridSensorsDataStreamRequest as PBReceiveMicrogridSensorsDataStreamRequest,
+)
+from frequenz.api.reporting.v1.reporting_pb2 import (
+    ReceiveMicrogridSensorsDataStreamResponse as PBReceiveMicrogridSensorsDataStreamResponse,
 )
 from frequenz.api.reporting.v1.reporting_pb2 import (
     ResamplingOptions as PBResamplingOptions,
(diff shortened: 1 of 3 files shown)


src/frequenz/client/reporting/_client.py
```diff
@@ -154,6 +163,74 @@ def __iter__(self) -> Iterator[MetricSample]:
...
+
+        Returns:
+            True if the batch contains no valid sensor data.
+        """
+        if not self._data_pb.sensors:
+            return True
+        for sensor in self._data_pb.sensors:
+            if not sensor.metric_samples and not sensor.states:
+                return True
+        return False
```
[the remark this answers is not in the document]
cwasicki: I think you can only check the first entry here.

src/frequenz/client/reporting/_client.py
```diff
@@ -154,6 +163,74 @@ def __iter__(self) -> Iterator[MetricSample]:
                         )
 
 
+@dataclass(frozen=True)
+class SensorsDataBatch:
```
cwasicki: Since this is duplicating a lot of code we have for components, I suggest we make the class more modular so that we can re-use the same code for both, components and sensors.

src/frequenz/client/reporting/_client.py
```diff
@@ -407,6 +484,189 @@ async def stream_method() -> (
...
+                yield response
+
+        broadcaster = GrpcStreamBroadcaster(
+            stream_name="microgrid-sensors-data-stream",
+            stream_method=stream_method,
+            transform=transform_response,
+            retry_strategy=None,
+        )
+
+        receiver = broadcaster.new_receiver()
```
cwasicki: Here we want to return the receiver directly (also for the component counterpart in future).

src/frequenz/client/reporting/_client.py
```diff
@@ -407,6 +484,189 @@ async def stream_method() -> (
...
+            for metric in metrics
+        ]
+
+        request = PBReceiveMicrogridSensorsDataStreamRequest(
+            microgrid_sensors=microgrid_sensors_pb,
+            metrics=metric_conns_pb,
+            filter=stream_filter,
+        )
+
+        def transform_response(
```
cwasicki, line 590: I wonder if it is possible to have the flattening loop inside this transform function so that we don't need the private batch function but directly get a receiver of `MetricSamples`.

src/frequenz/client/reporting/_client.py
```diff
@@ -407,6 +484,189 @@ async def stream_method() -> (
         async for data in receiver:
             yield data
 
+    # pylint: disable=too-many-arguments
+    async def list_single_sensor_data(
```
cwasicki: The `list` in the method name is outdated since it can also stream now. I think `receive` would be better.

src/frequenz/client/reporting/_client.py
```diff
@@ -407,6 +484,189 @@ async def stream_method() -> (
...
+        self,
+        *,
+        microgrid_id: int,
+        sensor_id: int,
+        metrics: Metric | list[Metric],
+        start_dt: datetime | None,
```
cwasicki, line 440: These arguments are also different from our other APIs, where we mostly use `start`/`end`. However, I guess we will change this soon to `{start,end}_time`. Better clarify before merging.

cwasicki (commented):
Since we are replicating a lot of code here I would be in favor of refactoring some of our initial code now. If it's very urgent to get in we could also merge this first, but then we should carefully look what we expose now to reduce breaking changes in future.

src/frequenz/client/reporting/_client.py
flora-hofmann-frequenz: Should I post this in the channel and then open a separate PR based on the consensus?

src/frequenz/client/reporting/_client.py
flora-hofmann-frequenz: I read up about this a little and would rather like to address it in a separate issue.

src/frequenz/client/reporting/_client.py
cwasicki: We discussed this yesterday and decided to go with `{start,end}_time`

cwasicki (approved):
Cool! LGTM

src/frequenz/client/reporting/_client.py
flora-hofmann-frequenz: Will add it in a small PR!

PR #176, opened 2025-05-06
Add sensor request to readme

flora-hofmann-frequenz:


README.md
```diff
@@ -70,6 +70,22 @@ data = [
...
+### Query metrics for a single microgrid and sensor:
+
+```python
+data = [
+    sample async for sample in
+    client.receive_single_sensor_data(
+        microgrid_id=1,
+        sensor_id=100,
+        metrics=[Metric.SENSOR_IRRADIANCE],
+        start_dt=datetime.fromisoformat("2024-05-01T00:00:00"),
```
cwasicki, line 82: We should first change these argument names to `{start,end}_time` before sending this out.

flora-hofmann-frequenz: This would be with the next release tough. Should I just change it for sensors & in this PR and then make a new release?

cwasicki: Good point, let's get this in, then change the arguments and make a new release with this and the receiver changes soon.

cwasicki (approved):

README.md
flora-hofmann-frequenz: Addressed in #178

PR #177, opened 2025-05-08
Update components retrival functions

flora-hofmann-frequenz:
Implemented comments from #169 for components.


src/frequenz/client/reporting/_client.py
```diff
@@ -343,7 +345,7 @@ async def _list_microgrid_components_data_batch(
             include_states: Whether to include the state data.
             include_bounds: Whether to include the bound data.
 
-        Yields:
+        Returns:
```
shsms: You might want to change the return type a receiver as well.

src/frequenz/client/reporting/_client.py
```diff
@@ -426,9 +428,7 @@ async def stream_method() -> (
             retry_strategy=None,
         )
 
-        receiver = broadcaster.new_receiver()
-        async for data in receiver:
-            yield data
+        return broadcaster.new_receiver()
```
shsms: This appears to be creating a new `GrpcStreamBroadcaster` everytime a new receiver is requested.  But that would create a new connection to the API service each time.  Instead you should hold on to the `broadcaster` instance and reuse it, like here:
- https://github.com/frequenz-floss/frequenz-client-electricity-trading-python/blob/a029d214ea7cdaf7f84ed72c7990546705e7a558/src/frequenz/client/electricity_trading/_client.py#L210-L215
- https://github.com/frequenz-floss/frequenz-client-electricity-trading-python/blob/a029d214ea7cdaf7f84ed72c7990546705e7a558/src/frequenz/client/electricity_trading/_client.py#L299-L322

shsms:
Hi Flora, just realized why this was complicated.  Christoph and I had solved this some time back.  I've made a PR with it into your branch: https://github.com/flora-hofmann-frequenz/frequenz-client-reporting-python/pull/1

shsms:
@flora-hofmann-frequenz I pushed to your branch directly because I was too lazy to create a new PR.

@cwasicki fyi

src/frequenz/client/reporting/_client.py
```diff
@@ -137,12 +137,12 @@ async def receive_single_component_data(
             include_states: Whether to include the state data.
             include_bounds: Whether to include the bound data.
 
-        Yields:
+        Returns:
```
cwasicki, line 155: Returns:
    A receiver of `MetricSamples`?

src/frequenz/client/reporting/_client.py
```diff
@@ -182,15 +178,15 @@ async def receive_microgrid_components_data(
...
```
cwasicki, line 194: See above

cwasicki (dismissed):
Just minor doc suggestion.

src/frequenz/client/reporting/_client.py
shsms: fixed.

src/frequenz/client/reporting/_client.py
shsms: fixed.

shsms:
I just realized we need to do the same `BatchUnrollReceiver` for the sensor data as well.  Let me do that as well quickly in this PR.

shsms:
Also did for the AggregatedDataStream, and some more simplification and cleanup.

cwasicki (approved):

PR #181, opened 2025-05-22
Add HMAC capabilities

florian-wagner-frequenz:
This adds the capability for HMAC based signing into the reporting client.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -32,7 +32,7 @@ dependencies = [
   "frequenz-client-common >= 0.3.0, < 0.4",
   "grpcio >=1.70.0, < 2",
   "protobuf >= 5.29.3, < 7",
-  "frequenz-client-base >= 0.8.0, < 0.11.0",
+  "frequenz-client-base >= 0.11.0, < 0.12.0",
 ]
 dynamic = ["version"]
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -156,6 +156,10 @@ disable = [
 
 [tool.pytest.ini_options]
 addopts = "-vv"
+testpaths = ["tests", "src"]
+asyncio_mode = "auto"
+asyncio_default_fixture_loop_scope = "function"
+required_plugins = ["pytest-asyncio", "pytest-mock"]
 filterwarnings = [
   "error",
   "once::DeprecationWarning",
@@ -164,10 +168,6 @@ filterwarnings = [
   # chars as this is a regex
   'ignore:Protobuf gencode version .*exactly one major version older.*:UserWarning',
 ]
-testpaths = ["tests", "src"]
-asyncio_mode = "auto"
-asyncio_default_fixture_loop_scope = "function"
-required_plugins = ["pytest-asyncio", "pytest-mock"]
 
 [tool.mypy]
 explicit_package_bases = true
--- a/src/frequenz/client/reporting/_client.py
+++ b/src/frequenz/client/reporting/_client.py
@@ -64,10 +64,13 @@ from ._types import (
 class ReportingApiClient(BaseApiClient[ReportingStub]):
     """A client for the Reporting service."""
 
+    # pylint: disable-next=too-many-arguments
     def __init__(
         self,
         server_url: str,
-        key: str | None = None,
+        *,
+        auth_key: str | None = None,
+        sign_secret: str | None = None,
         connect: bool = True,
         channel_defaults: ChannelOptions = ChannelOptions(),  # default options
     ) -> None:
@@ -75,7 +78,8 @@ class ReportingApiClient(BaseApiClient[ReportingStub]):
 
         Args:
             server_url: The URL of the Reporting service.
-            key: The API key for the authorization.
+            auth_key: The API key for the authorization.
+            sign_secret: The secret to use for HMAC signing the message
             connect: Whether to connect to the server immediately.
             channel_defaults: The default channel options.
         """
@@ -84,6 +88,8 @@ class ReportingApiClient(BaseApiClient[ReportingStub]):
             ReportingStub,
             connect=connect,
             channel_defaults=channel_defaults,
+            auth_key=auth_key,
+            sign_secret=sign_secret,
         )
 
         self._components_data_streams: dict[
@@ -128,8 +134,6 @@ class ReportingApiClient(BaseApiClient[ReportingStub]):
             ],
             GrpcStreamBroadcaster[PBAggregatedStreamResponse, MetricSample],
         ] = {}
-
-        self._metadata = (("key", key),) if key else ()
 
     @property
     def stub(self) -> ReportingStub:
@@ -309,7 +313,7 @@ class ReportingApiClient(BaseApiClient[ReportingStub]):
                 AsyncIterable[PBReceiveMicrogridComponentsDataStreamResponse]
             ):
                 call_iterator = self.stub.ReceiveMicrogridComponentsDataStream(
-                    request, metadata=self._metadata
+                    request,
                 )
                 return cast(
                     AsyncIterable[PBReceiveMicrogridComponentsDataStreamResponse],
@@ -493,9 +497,7 @@ class ReportingApiClient(BaseApiClient[ReportingStub]):
             def stream_method() -> (
                 AsyncIterable[PBReceiveMicrogridSensorsDataStreamResponse]
             ):
-                call_iterator = self.stub.ReceiveMicrogridSensorsDataStream(
-                    request, metadata=self._metadata
-                )
+                call_iterator = self.stub.ReceiveMicrogridSensorsDataStream(request)
                 return cast(
                     AsyncIterable[PBReceiveMicrogridSensorsDataStreamResponse],
                     call_iterator,
@@ -588,7 +590,7 @@ class ReportingApiClient(BaseApiClient[ReportingStub]):
             def stream_method() -> AsyncIterable[PBAggregatedStreamResponse]:
                 call_iterator = (
                     self.stub.ReceiveAggregatedMicrogridComponentsDataStream(
-                        request, metadata=self._metadata
+                        request,
                     )
                 )
 
--- a/src/frequenz/client/reporting/cli/__main__.py
+++ b/src/frequenz/client/reporting/cli/__main__.py
@@ -79,9 +79,15 @@ def main() -> None:
         "--format", choices=["iter", "csv", "dict"], help="Output format", default="csv"
     )
     parser.add_argument(
-        "--key",
+        "--auth_key",
         type=str,
         help="API key",
+        default=None,
+    )
+    parser.add_argument(
+        "--sign_secret",
+        type=str,
+        help="The secret to use for generating HMAC signatures",
         default=None,
     )
     args = parser.parse_args()
@@ -96,8 +102,9 @@ def main() -> None:
             states=args.states,
             bounds=args.bounds,
             service_address=args.url,
-            key=args.key,
+            auth_key=args.key,
             fmt=args.format,
+            sign_secret=args.sign_secret,
         )
     )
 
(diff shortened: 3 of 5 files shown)


src/frequenz/client/reporting/cli/__main__.py
```diff
@@ -116,6 +126,7 @@ async def run(  # noqa: DOC502
     bounds: bool,
     service_address: str,
     key: str,
+    secret: str,
```
llucax: Maybe call this `signing_secret` for extra clarity?

src/frequenz/client/reporting/cli/__main__.py
```diff
@@ -130,13 +141,19 @@ async def run(  # noqa: DOC502
...
     Raises:
         ValueError: if output format is invalid
     """
-    client = ReportingApiClient(service_address, key)
+    auth = AuthenticationOptions(key)
+    sign = SigningOptions(secret)
+
+    channel_options = ChannelOptions(auth=auth, sign=sign)
+
+    client = ReportingApiClient(service_address, key, channel_defaults=channel_options)
```
llucax: This looks pretty inconsistent, IMHO this should be done inside the client class, and the signing secret should be taken as a separate argument, like with `key`. The options passed here are **default options**. There is also an overlap, because [the client is doing its own `key` management](https://github.com/frequenz-floss/frequenz-client-reporting-python/blob/v0.x.x/src/frequenz/client/reporting/_client.py#L227), which if it doesn't collide with the new auth interceptor, it will overlap.

Looking at the [`ReportingApiClient.__init__`](https://github.com/frequenz-floss/frequenz-client-reporting-python/blob/c60b4ab6b374bd97d240f1eb5b2c9a9f77de95db/src/frequenz/client/reporting/_client.py#L200-L227), I feel like the current approach we taken with the auth key and sign secret is not consistent with the default channel options :-/

All channel options were designed to be mainly set from the `server_url`, and the `channel_defauls` is supposed to just pass defaults, but the real end values are configured in the `server_url` if different to the default. But we don't support setting the auth key and sign secret. This maybe makes sense in a way, as it probably makes more sense to pass secrets in separate env vars in the future, instead of treating the whole `server_url` as a secret. But it doesn't make sense to use a *default secret* that is not `None`, as either you want to use a secret, in which case you need to specify one, or you don't, but there is no *default secret* one can provide.

I think we should probably change the base client to pass secrets outside the `default_options` to the client. Which means there are no more `sign` and `auth` option in channel options. So the base client ctor will take a `auth_key` and `sign_secret` as explicit arguments defaulting to `None`. So you either pass it or not, but there is no *default*. Sorry @florian-wagner-frequenz :pray:, I think this was your original approach and I suggested using the channel options. I think when I proposed it the idea was also to add eventually support to setting this via env vars, which is more reasonable, but I also agree reading env vars deep in some library, specially if they are secrets, is a bad idea.

A for a temporary solution I suggest changing the current [`ReportingApiClient.__init__`](https://github.com/frequenz-floss/frequenz-client-reporting-python/blob/c60b4ab6b374bd97d240f1eb5b2c9a9f77de95db/src/frequenz/client/reporting/_client.py#L200-L227) to something like:

```py
class ReportingApiClient(BaseApiClient[ReportingStub]):
    """A client for the Reporting service."""


    def __init__(
        self,
        server_url: str,
        auth_key: str | None = None,
        sign_secret: str | None = None,
        connect: bool = True,
        channel_defaults: ChannelOptions = ChannelOptions(),  # default options
    ) -> None:
        channel_defaults = channel_defaults.replace(sign=SigningOptions(secret=sign_secret))
        channel_defaults = channel_defaults.replace(auth=AuthenticationOptions(api_key=auth_key))
        super().__init__(
            server_url,
            ReportingStub,
            connect=connect,
            channel_defaults=channel_defaults,
        )
```

pyproject.toml
```diff
@@ -155,7 +155,7 @@ disable = [
 ]
 
 [tool.pytest.ini_options]
-addopts = "-W=all -Werror -Wdefault::DeprecationWarning -Wdefault::PendingDeprecationWarning -vv"
+addopts = "-W=all -Wdefault::DeprecationWarning -Wdefault::PendingDeprecationWarning -vv"
```
llucax: `-Werror` should be kept

src/frequenz/client/reporting/cli/__main__.py
florian-wagner-frequenz: Gladly

pyproject.toml
florian-wagner-frequenz: okay, but I think then I might need to put the specific warning back in. This is the fix Flora applied.

llucax: oh, really, I didn't realized that, I guess I missed that change as there were other changes too in that line. Then something went wrong. @flora-hofmann-frequenz :-/

Removing `-Werror` means we don't generally treat warnings as errors anymore. I will have a look.

src/frequenz/client/reporting/cli/__main__.py
florian-wagner-frequenz: Okay, I will do the temporary fix first and see how far I get with doing it properly in the base client. I agree with your arguments.
For me the core issue is:
- `key` and `secret` are not optional (as the service won't respond), so shouldn't live in the `channel_defaults` (as we can not conjure them from thin air
- We don't want to pull them from an environment variable deep in the library (compounding the previous issue)
- We want them to stay secret, so we don't want them in the server_url

So the targeted interface for the clients/apps stays to pass via command line argument (possibly from environment variable).
This of course pushes the burden to app maintainers, but I think that is unavoidable unless we want the "pull from environment variable in library" solution.

pyproject.toml
llucax: OK, I found a fix, and while at it, tried a few improvements, with these options we don't need to add the `conftest.py` file at all!

```toml
[tool.pytest.ini_options]
addopts = "-vv"
filterwarnings = [
  "error",
  "once::DeprecationWarning",
  "once::PendingDeprecationWarning",
  # We use a raw string (single quote) to avoid the need to escape special
  # chars as this is a regex
  'ignore:Protobuf gencode version .*exactly one major version older.*:UserWarning',
]
```

I think it is better to complete ignore the protobuf warnings, and also not care about the reported version (so we don't need to update the ignore rule when we bump the protobuf version), we know we are doing this, so it only adds noise to the tests to show it. We'll get a proper error if we go beyond the supported range anyways.

Also show deprecations only once, we don't need to get spammed if a deprecated symbol is used more than once.

I will add these improvements to `repo-config`.

llucax: * https://github.com/frequenz-floss/frequenz-repo-config-python/issues/407

src/frequenz/client/reporting/cli/__main__.py
florian-wagner-frequenz: I ended up going the long way, but now we can use it as an example to others.

src/frequenz/client/reporting/cli/__main__.py
florian-wagner-frequenz: I called it `sign_secret` to be in line with the base client.

src/frequenz/client/reporting/cli/__main__.py
```diff
@@ -130,13 +141,19 @@ async def run(  # noqa: DOC502
...
-        key: API key
+        key: API key (per customer)
+        secret: cryptographic secret of the service (per API)
         fmt: output format
 
     Raises:
         ValueError: if output format is invalid
     """
-    client = ReportingApiClient(service_address, key)
+    auth = AuthenticationOptions(key)
```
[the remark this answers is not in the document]
florian-wagner-frequenz: This is no longer needed, we just pass through the secret to the BaseApiClient

pyproject.toml
```diff
@@ -155,11 +155,19 @@ disable = [
...
 asyncio_default_fixture_loop_scope = "function"
 required_plugins = ["pytest-asyncio", "pytest-mock"]
+filterwarnings = [
+  "error",
+  "once::DeprecationWarning",
+  "once::PendingDeprecationWarning",
+  # We use a raw string (single quote) to avoid the need to escape special
+  # chars as this is a regex
+  'ignore:Protobuf gencode version .*exactly one major version older.*:UserWarning',
+]
```
llucax: Nitpick: Could you split this into a different commit?

src/frequenz/client/reporting/_client.py
```diff
@@ -200,10 +200,12 @@ def sample(self) -> MetricSample:
 class ReportingApiClient(BaseApiClient[ReportingStub]):
     """A client for the Reporting service."""
 
+    # pylint: disable=too-many-arguments, too-many-positional-arguments
```
llucax: If you use `disable=` in its own line, that will disable the check for the rest of the file, not only for the next line, if you want to disable it only for the next line you need to use `disable-next=`.

```suggestion
    # pylint: disable-next=too-many-arguments, too-many-positional-arguments
```

Also I think in this case `pylint` has a point, and you should use positional-only arguments, specially for strings/bool that are really hard to read without a keyword on the call site.

src/frequenz/client/reporting/_client.py
```diff
@@ -200,10 +200,12 @@ def sample(self) -> MetricSample:
...
     def __init__(
         self,
         server_url: str,
         key: str | None = None,
```
llucax:
```suggestion
        *,
        key: str | None = None,
```

src/frequenz/client/reporting/cli/__main__.py
llucax: Great, thanks!

src/frequenz/client/reporting/_client.py
```diff
@@ -220,6 +223,8 @@ def __init__(
             ReportingStub,
             connect=connect,
             channel_defaults=channel_defaults,
+            auth_key=key,
```
llucax: You still need to remove the ad-hoc code in this class to deal with the auth key (`self._metadata`).

src/frequenz/client/reporting/_client.py
florian-wagner-frequenz: agreed

src/frequenz/client/reporting/_client.py
florian-wagner-frequenz: Again, me being to meek about breaking the interface, but I am already doing it anyways.

llucax: We are at 0.x, so...

BREAK ALL THE THINGS!

src/frequenz/client/reporting/_client.py
```diff
@@ -64,10 +64,13 @@
 class ReportingApiClient(BaseApiClient[ReportingStub]):
     """A client for the Reporting service."""
 
+    # pylint: disable-next=too-many-arguments
     def __init__(
         self,
         server_url: str,
+        *,
         key: str | None = None,
```
llucax: Sorry for the delay and not seeing this before, but it would also be nice to have consistent client interfaces, so I would also rename this to `auth_key`:
```suggestion
        auth_key: str | None = None,
```
:grimacing:

src/frequenz/client/reporting/cli/__main__.py
```diff
@@ -84,6 +84,12 @@ def main() -> None:
         help="API key",
```
llucax, line 84: I would also rename `--key` to `--auth_key` (or add an alias if you prefer to avoid breaking changes, printing a deprecation message if `--key` is used).

llucax (dismissed):
Approving as the `key` rename is really not related to this PR, so it can be done in a separate PR.

src/frequenz/client/reporting/cli/__main__.py
florian-wagner-frequenz: I shall break all the eggs

llucax (approved):

src/frequenz/client/reporting/_client.py
```diff
@@ -64,18 +64,22 @@
 class ReportingApiClient(BaseApiClient[ReportingStub]):
     """A client for the Reporting service."""
 
+    # pylint: disable-next=too-many-arguments
     def __init__(
         self,
         server_url: str,
-        key: str | None = None,
+        *,
+        auth_key: str | None = None,
```
cwasicki, line 72: Could you please update the README w.r.t. these interface changes.

PR #187, opened 2025-06-05
Update proto warnings

flora-hofmann-frequenz:


RELEASE_NOTES.md
```diff
@@ -16,4 +16,4 @@
 
 ## Bug Fixes
 
-<!-- Here goes notable bug fixes that are worth a special mention or explanation -->
+* Updated display of protobuf version warnings
```
llucax: This really shouldn't go to the release notes, it is internal (testing), so it doesn't affect the **users** of this library.

pyproject.toml
```diff
@@ -30,8 +30,6 @@ dependencies = [
   "grpc-stubs == 1.53.0.6",
   "frequenz-api-reporting >= 0.7.0, < 0.8",
   "frequenz-client-common >= 0.3.0, < 0.4",
-  "grpcio >=1.70.0, < 2",
-  "protobuf >= 5.29.3, < 7",
```
llucax: This should probably be separated in a different commit or at least be clarified in the commit message.

But you should probably not remove the `protobuf` dependency, as [you are using it in your code](https://github.com/frequenz-floss/frequenz-client-reporting-python/blob/5f815fd0057a7b5c85e22890a3cb57d5564b7338/src/frequenz/client/reporting/_client.py#L53).

flora-hofmann-frequenz: Oh true, forgot about the `Timestamps`... I will add a commit and bump to `"grpcio >= 1.72.1, < 2", "protobuf >= 6.31.1, < 8",` or should I rather not add this bump to this PR?

llucax (approved):

PR #189, opened 2025-06-09
Take out async for sensors

flora-hofmann-frequenz:
Think we forgot to take out 'async' when refactoring in #177

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -14,4 +14,4 @@
 
 ## Bug Fixes
 
-<!-- Here goes notable bug fixes that are worth a special mention or explanation -->
+* Take out 'async' from  'receive_microgrid_sensors_data' and 'receive_single_sensor_data'.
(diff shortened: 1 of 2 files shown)


shsms:
Oops, looks like we need to remove it from the aggregated data method as well.

flora-hofmann-frequenz:
> Oops, looks like we need to remove it from the aggregated data method as well.

I would like to do that in a separate PR.

I would also have to touch `_batch_unroll_receiver` to add the Receiver for aggregated data, no? Or would I leave it as an iterator?

shsms:
> I would also have to touch _batch_unroll_receiver to add the Receiver for aggregated data, no? Or would I leave it as an iterator?

oh damn, I did that, but pushed to my branch and not to yours, even though I wrote in your PR that I had done it, so it didn't get merged. 🤦🏽 

I'll make a PR with the remaining commits now.

shsms (approved):

flora-hofmann-frequenz:
> oh damn, I did that, but pushed to my branch and not to yours, even though I wrote in your PR that I had done it, so it didn't get merged. 🤦🏽
> 
> I'll make a PR with the remaining commits now.

Great, thanks a lot @shsms

PR #192, opened 2025-06-16
Update Readme for HMAC

florian-wagner-frequenz:


README.md
```diff
@@ -158,6 +162,7 @@ microgrid component data from the reporting API.
 reporting-cli \
     --url localhost:4711 \
     --key=$(<api_key.txt)
```
cwasicki: This has also changed

README.md
```diff
@@ -39,14 +39,18 @@ pip install frequenz-client-reporting==$VERSION
...
 ```python
 from datetime import datetime, timedelta
+import os
 
 from frequenz.client.common.metric import Metric
 from frequenz.client.reporting import ReportingApiClient
 
 # Change server address
 SERVER_URL = "grpc://replace-this-with-your-server-url:port"
 API_KEY = open('api_key.txt').read().strip()
```
cwasicki: I suggest we also get this from the env to make it consistent with the way we pass in the the secret.

README.md
florian-wagner-frequenz: Ah, right. I checked at the wrong place and there is actually a bug here I need to fix :)

README.md
florian-wagner-frequenz: Gladly

RELEASE_NOTES.md
```diff
@@ -6,13 +6,14 @@
...
 
-<!-- Here goes notes on how to upgrade from previous versions, including deprecations and what they should be replaced with -->
+* The `key` parameter of the client has been renamed to `auth_key`
 
 ## New Features
 
 * Add HMAC generation capabilities.
-    * The new CLI option "key" can be used to provide the server's key.
-    * The client itself now has a "key" argument in the constructor.
+    * The new CLI option "sign_secret" can be used to provide the server's key.
```
Marenz: It's confusing to read that the key should go into "sign_secret" when there is also "auth_key" formerly key :sweat_smile:

cwasicki (dismissed):

RELEASE_NOTES.md
florian-wagner-frequenz: okay, that one is definitely confusing and should be fixed in the changelog

florian-wagner-frequenz: Fixed it to be clear that it's the server's secret (aka API secret)

Marenz (approved):

PR #203, opened 2025-07-08
Add support for AggregatedMetricValue

micaebe:
Adds support to read `AggregatedMetricValue` (besides `SimpleMetricValue`) inside `GenericDataBatch`.

### DIFF
--- a/src/frequenz/client/reporting/_types.py
+++ b/src/frequenz/client/reporting/_types.py
@@ -73,8 +73,8 @@ class GenericDataBatch:
     def __iter__(self) -> Iterator[MetricSample]:
         """Get generator that iterates over all values in the batch.
 
-        Note: So far only `SimpleMetricSample` in the `MetricSampleVariant`
-        message is supported.
+        Note: `SimpleMetricValue` and `AggregatedMetricValue` in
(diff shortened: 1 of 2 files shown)


src/frequenz/client/reporting/_types.py
```diff
@@ -93,11 +93,13 @@ def __iter__(self) -> Iterator[MetricSample]:
...
-                    if sample.value.HasField("simple_metric")
-                    else math.nan
-                )
+
+                if sample.value.HasField("simple_metric"):
+                    value = sample.value.simple_metric.value
+                elif sample.value.HasField("aggregated_metric"):
+                    value = sample.value.aggregated_metric.avg_value
+                else:
+                    value = math.nan
```
cwasicki, line 102: This part could be moved into a separate method to avoid the CI failure. In fact this whole class is on the list for refactoring anyway, and this would be a reasonable change.

cwasicki (commented):
Thanks a lot for the fix! LGTM, small comment to avoid the CI failure. Also needs signing of the commits.

flora-hofmann-frequenz:
Addressed in PR #204

PR #219, opened 2025-08-19
Edit README

merlin-esser-frequenz:
Add more details for setting up and using the API

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,6 @@ A Python client for interacting with the Frequenz Reporting API to efficiently r
 > **Who should use this?**
 > This client is for developers building applications on top of Frequenz's platform who need structured access to historical component or sensor data via Python or CLI.
 
-
 ## Supported Platforms
 
 The following platforms are officially supported (tested):
@@ -27,7 +26,7 @@ check out the [Contributing Guide](CONTRIBUTING.md).
 
 ## Getting Started
 
-To get started, follow the steps below to set up your development environment and run the client.
+To get started, create a directory for your project and follow the steps below to set up your development environment and run the client.
 
 ### 1. Install VS Code
 
@@ -45,22 +44,27 @@ These enable Python development and interactive notebooks within VS Code.
 It is recommended to use a **virtual environment** for isolating dependencies.
 When using a virtual environment, you create something like a sandbox or containerized workspace that isolates dependencies and packages so they don’t conflict with the system or other projects. After creating the virtual environment, you can install all packages you need for the current project. To set up an initial virtual environment for using the Reporting API, go through the following steps:
 
-### Linux / macOS
+#### Linux / macOS
 
 ```bash
+### Go to your projects directory, open a terminal and run the following commands
+
 # Create a virtual environment
 python3 -m venv .venv
 
 # Activate the environment
-source .venv/bin/activate
+source .venv/bin/activate # when using bash
+source .venv/bin/activate.fish # when using fish
 
 # Install the version of the reporting client you want to use
 pip install frequenz-client-reporting==0.18.0
 
-# Install python-dotenv --> Used to load environment variables from a `.env` file into the projects environment
+# Install python-dotenv
+# --> Used to load environment variables from a `.env` file into the projects environment
 pip install python-dotenv
 
-# Install ipkernel --> Python execution backend for Jupyter
+# Install ipkernel
+# --> Python execution backend for Jupyter
 pip install ipkernel
 
 # Register the virtual environment as a kernel
@@ -68,6 +72,41 @@ python -m ipykernel install --user --name=myenv --display-name "Python (myenv)"
 
 # Install pandas
 pip install pandas
+
+# Deactivate the environment
+deactivate
+```
+
+#### Windows (PowerShell)
+
+```powershell
+### Go to your projects directory, open PowerShell and run the following commands
+
+# Create a virtual environment
+python -m venv .venv
+
+# Activate the environment
+.venv\Scripts\Activate.ps1
+
+# Install the version of the reporting client you want to use
+pip install frequenz-client-reporting==0.18.0
+
+# Install python-dotenv
+# --> Used to load environment variables from a `.env` file into the project's environment
+pip install python-dotenv
+
+# Install ipykernel
+# --> Python execution backend for Jupyter
+pip install ipykernel
+
+# Register the virtual environment as a kernel
+python -m ipykernel install --user --name=myenv --display-name "Python (myenv)"
+
+# Install pandas
+pip install pandas
+
+# Deactivate the environment
+deactivate
 ```
 
 ## Create API Keys via Kuiper
@@ -76,7 +115,8 @@ To access the API, a key has to be created via Kuiper and stored locally on your
 
 ### 1 create .env file
 
-When creating the API key, you will get two keys. One is the API key itself and the other one is the API secret. Create a textfile named ".env" in your project folder where you can store the API key and the API secret.
+When creating the API key, you will get two keys. One is the API key itself and the other one is the API secret. Go to your projects directory and create a textfile named ".env" in your project folder where you can store the API key and the API secret.
+Prepare the textfile as follows:
 
 ```bash
 # .env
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -25,42 +24,155 @@ The following platforms are officially supported (tested):
...
 check out the [Contributing Guide](CONTRIBUTING.md).
 
+## Getting Started
 
-## Usage
+To get started, create a directory for your project and follow the steps below to set up your development environment and run the client.
 
-Please also refer to source of the [CLI tool](https://github.com/frequenz-floss/frequenz-client-reporting-python/blob/v0.x.x/src/frequenz/client/reporting/cli/__main__.py)
-for a practical example of how to use the client.
+### 1. Install VS Code
```
cwasicki, line 31: This is not related to the reporting API, but a general recommendation how to set up a working environment. If we want to document this, I suggest to put it into another place than the reporting client README and link to it from here. 

Apart from that, this sounds like vscode is required for the reporting API. It is not, and IMO it's overkill to recommend it.

README.md
```diff
@@ -25,42 +24,155 @@ The following platforms are officially supported (tested):
...
+Download and install **Visual Studio Code** from the [official website](https://code.visualstudio.com/).
+
+Once installed, add the following extensions:
+
+- [Python extension](https://marketplace.visualstudio.com/items?itemName=ms-python.python)  
+- [Jupyter extension](https://marketplace.visualstudio.com/items?itemName=ms-toolsai.jupyter)  
+
+These enable Python development and interactive notebooks within VS Code.
+
+### 2. Set Up a Virtual Environment
```
cwasicki, line 42: Same for this section, not only relevant for reporting API. Do we have instructions on environments already somewhere documented @llucax ?

cwasicki: The problem with documenting it here is that we will have to maintain similar sections in all client repositories.

README.md
```diff
@@ -25,42 +24,155 @@ The following platforms are officially supported (tested):
...
+# Activate the environment
+source .venv/bin/activate # when using bash
+source .venv/bin/activate.fish # when using fish
+
+# Install the version of the reporting client you want to use
+pip install frequenz-client-reporting==0.18.0
+
+# Install python-dotenv
+# --> Used to load environment variables from a `.env` file into the projects environment
+pip install python-dotenv
```
cwasicki, line 64: This should be optional

README.md
```diff
@@ -25,42 +24,155 @@ The following platforms are officially supported (tested):
...
+
+# Install ipkernel
+# --> Python execution backend for Jupyter
+pip install ipkernel
```
cwasicki, line 68: Also optional. Moreover I think there is a typo in the package name.

README.md
```diff
@@ -25,42 +24,155 @@ The following platforms are officially supported (tested):
...
+
+# Register the virtual environment as a kernel
+python -m ipykernel install --user --name=myenv --display-name "Python (myenv)"
```
cwasicki, line 71: All of these things are optional and not required for the reporting API.

README.md
```diff
@@ -25,42 +24,155 @@ The following platforms are officially supported (tested):
...
 ```
 
+### 2. Create API key via Kuiper
+
+Log into your Kuiper account and click on your email-address on the bottom left. Then continue with `API keys` and `Create API key`. The created API key will be shown in the UI. This is the only time, the API key will be shown to you. whenever you loose it, you have to create a new one.
+Copy the API key and the API secret and add it to the `.env` file.
+
+```bash
+# .env
+REPORTING_API_AUTH_KEY=EtQurK8LXA8vmEd4M6DqxeNp
```
cwasicki, line 134: I hope this is not what it looks like.

README.md
matthias-wende-frequenz, line 31: I don't think we shouldn't advertise specific editors in this library.

README.md
```diff
@@ -25,42 +24,155 @@ The following platforms are officially supported (tested):
...
+
+> **Security Note:**
 > Always keep your authentication key and signing secret secure. Do not hard-code them in your source code or share them publicly.
 
 ```python
+# Import basic packages
+from dotenv import load_dotenv
 from datetime import datetime, timedelta
 import os
+import pandas as pd
```
cwasicki, line 164: Why this import?

README.md
```diff
@@ -25,42 +24,155 @@ The following platforms are officially supported (tested):
...
+These enable Python development and interactive notebooks within VS Code.
+
+### 2. Set Up a Virtual Environment
+
+It is recommended to use a **virtual environment** for isolating dependencies.
+When using a virtual environment, you create something like a sandbox or containerized workspace that isolates dependencies and packages so they don’t conflict with the system or other projects. After creating the virtual environment, you can install all packages you need for the current project. To set up an initial virtual environment for using the Reporting API, go through the following steps:
+
+#### Linux / macOS
+
+```bash
```
matthias-wende-frequenz, line 49: In general I like the idea of having copy paste friendly instructions. Maybe we should only focus on the reporting part in this case and skip the notebook part. I also suggest to create a separate manual page for dealing with notebooks and keep the `README` at a bare minimum.

README.md
```diff
@@ -25,42 +24,155 @@ The following platforms are officially supported (tested):
...
+
+When creating the API key, you will get two keys. One is the API key itself and the other one is the API secret. Go to your projects directory and create a textfile named `.env` where you can store the API key and the API secret.
+Prepare the textfile as follows:
 
 ```bash
-# Choose the version you want to install
-VERSION=0.18.0
-pip install frequenz-client-reporting==$VERSION
+# .env
+REPORTING_API_AUTH_KEY=
```
cwasicki, line 123: Probably good to follow our standard naming `REPORTING_API_{URL,KEY,SECRET}`

matthias-wende-frequenz (changes requested):
I'm unsure if add specific use cases to the readme is what we want. You should also squash your commits or give each of them a meaningful commit message.

cwasicki (changes requested):
I don't think these recommendations should go into the reporting client README, because they are 1) optional recommendations not required for the reporting client to run, 2) not specific to the reporting client but potentially useful for other python clients or our python tools in general.

README.md
matthias-wende-frequenz: You can be relieved it's not an active key!

README.md
thomas-nicolai-frequenz: I agreee with @matthias-wende-frequenz

README.md
thomas-nicolai-frequenz: Yes, please.

README.md
thomas-nicolai-frequenz: Please remove the key no matter if valid or not and replace with just a template parameter like <here-goes-your-key>. Surely, you can find a suggestion from the GPTs of this world ;-)

README.md
merlin-esser-frequenz: to use pandas data frames

README.md
merlin-esser-frequenz: @cwasicki this is copied from the current README

merlin-esser-frequenz:
Thanks for the review!

From what I read, I realized, that I didnt really understand, what the README is supposed to be used for.
From my thinking, I wanted to collect a bunch of information for Customers, who are not that familiar with things like that, and give them a point to start.

So I guess a pdf where I summarize these steps and actively highlight, that these are just suggestions to start and that the API can be used in a lot of different other ways, makes definetly more sense.

After talking to the Customer, I realized, that they are only little steps away from really using the API to create their own notebooks, and for that I wanted to provide some help.

matthias-wende-frequenz:
> So I guess a pdf where I summarize these steps and actively highlight, that these are just suggestions to start and that the API can be used in a lot of different other ways, makes definetly more sense.
> 
> After talking to the Customer, I realized, that they are only little steps away from really using the API to create their own notebooks, and for that I wanted to provide some help.

This repository has an auto rendered documentation which is in sync with the code. That's the place, where all of the above can be placed, for instance as tutorials or simple or advanced usage examples.

https://frequenz-floss.github.io/frequenz-client-reporting-python/

README.md
llucax: * https://github.com/frequenz-floss/frequenz-repo-config-python/issues/266

Now we have https://github.com/frequenz-floss/docs/, so we can put common docs here and leave only project-specific stuff for the project. I think we still need some time to sort this out properly, but I have a rough idea of how we should move forward.

README.md
llucax: Agreed.

README.md
llucax: Also this applies to most if not all projects, so probably something to be discussed here:

* https://github.com/frequenz-floss/frequenz-repo-config-python/issues/266

README.md
```diff
@@ -25,42 +24,155 @@ The following platforms are officially supported (tested):
...
+
+```bash
+### Go to your projects directory, open a terminal and run the following commands
+
+# Create a virtual environment
+python3 -m venv .venv
+
+# Activate the environment
+source .venv/bin/activate # when using bash
+source .venv/bin/activate.fish # when using fish
```
llucax, line 57: I think we should stick with one example environment and let users with more niche configuration figure things out for themselves (i.e. skip fish and mention at the start of the readme that we assume `bash` as the shell), otherwise we need to maintain all possible user configuration in our docs, which doesn't make a lot of sense.

Again something for:

* https://github.com/frequenz-floss/frequenz-repo-config-python/issues/266

README.md
llucax: And I would remove it, we shouldn't be so opinionated on how users should run stuff. We could provide some separate guide for a newbie with some opinionated suggested configuration and IDE, but that should be definitely not part of the README.

README.md
llucax: I don't even know what this is 😆

README.md
```diff
@@ -25,42 +24,155 @@ The following platforms are officially supported (tested):
...
+pip install ipkernel
+
+# Register the virtual environment as a kernel
+python -m ipykernel install --user --name=myenv --display-name "Python (myenv)"
+
+# Install pandas
+pip install pandas
+
+# Deactivate the environment
+deactivate
```
llucax, line 77: Why deactivating the env? If users do this then nothing works.

README.md
```diff
@@ -25,42 +24,155 @@ The following platforms are officially supported (tested):
...
+```
+
+#### Windows (PowerShell)
```
llucax, line 80: We are not officially supporting Windows yet, so I wouldn't include a section on how to install in a platform we don't officially support. Maybe we should start supporting Windows, but that's a separate discussion. Again, all of this is pretty general, so we probably should include it in docs for all projects.

README.md
```diff
@@ -25,42 +24,155 @@ The following platforms are officially supported (tested):
...
+
+# Deactivate the environment
+deactivate
+```
+
+## API Keys
+
+To access the API, a key has to be created via Kuiper and stored locally on your computer. For that go through the following steps:
+
+### 1. create .env file
```
llucax, line 116: This is also opinionated, creating an `.env` file is not necessary, there are many other ways to define an env var.

README.md
llucax: Since we introduced signing keys we are trying to use these names as standard:

* `XXX_API_URL`
* `XXX_API_AUTH_KEY`
* `XXX_API_SIGN_SECRET`

as using plain `KEY` and `SECRET` could be ambiguous.

llucax (commented):
All in all, I think this is useful, but it is definitely not for a README.

BUT if you split what you created and add it to `docs/tutorials/basic.md`, then I guess we get a nice tutorial for newbies.

README.md
llucax: In the context of making this a separate tutorial for newbies, we could even leave this section with a big warning saying that Windows is not YET officially supported.

thomas-nicolai-frequenz: Well it might discourage some people start trying to use it on Windows and hence they might solve some issues for us on the way.

llucax:
Here is an example on how to convert this into a tutorial:

* https://github.com/merlin-esser-frequenz/frequenz-client-reporting-python/pull/1

llucax:
Closing in favor of #220.

PR #240, opened 2025-11-27
Update to reporting API v1alpha10

cwasicki:
The reporting API update includes the major update to common API v1alpha8. This also requires an update to the new metrics package in the client-common.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -2,11 +2,12 @@
 
 ## Summary
 
-<!-- Here goes a general summary of what this release is about -->
+> **Warning:** This client is compatible *only* with Reporting API `v1alpha10` or later.
+> Using with services that use an older API version **will cause failures**.
 
 ## Upgrading
 
-<!-- Here goes notes on how to upgrade from previous versions, including deprecations and what they should be replaced with -->
+* Breaking change to reporting API v1alpha10.
 
 ## New Features
 
--- a/src/frequenz/client/reporting/_types.py
+++ b/src/frequenz/client/reporting/_types.py
@@ -20,7 +20,7 @@ from frequenz.api.reporting.v1.reporting_pb2 import (
 )
 
 # pylint: enable=no-name-in-module
-from frequenz.client.common.metric import Metric
+from frequenz.client.common.metrics import Metric
 
 
 class MetricSample(NamedTuple):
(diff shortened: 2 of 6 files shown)


stefan-brus-frequenz (dismissed):
Not too well versed in pythonics but the PR looks fairly straightforward to me. LGTM

src/frequenz/client/reporting/_types.py
```diff
@@ -92,8 +92,8 @@ def __iter__(self) -> Iterator[MetricSample]:
         for item in items:
             cid = getattr(item, self.id_attr)
             for sample in getattr(item, "metric_samples", []):
-                ts = sample.sampled_at.ToDatetime().replace(tzinfo=timezone.utc)
-                met = Metric.from_proto(sample.metric).name
+                ts = sample.sample_time.ToDatetime().replace(tzinfo=timezone.utc)
+                met = Metric(sample.metric).name
```
llucax: You can also import from `frequenz.client.common.proto` and use:

```suggestion
                ts = datetime_from_proto(sample.sample_time)
                met = enum_from_proto(Metric, sample.metric).name
```

The datetime conversion is the same, it only avoids the mistake of forgetting the timezone. But when using `enum_from_proto()` we could support multiple api-common versions in the future, maybe. And it can handle when the proto added a new metric but the client doesn't know about it yet. In this case, your code will raise an exception.

Same for other conversions like these.

cwasicki: Good idea, updated.

llucax: I see you use `allow_invalid=False`, which also will raise.

I guess it is on purpose, and I see maybe changing it now might not be trivial, but long term I think it would be better to keep clients forward-compatible by allowing the server returning some values we don't know about yet instead of raising an exception.

This will allow users to still get some metric the client doesn't support yet, if for some reason they are stuck with an old client version.

RELEASE_NOTES.md
```diff
@@ -2,11 +2,12 @@
...
 ## Summary
 
-<!-- Here goes a general summary of what this release is about -->
+> **Warning:** This client is compatible *only* with Reporting API `v1alpha10` or later.
+> Using with services that use an older API version **will cause failures**.
 
 ## Upgrading
 
-<!-- Here goes notes on how to upgrade from previous versions, including deprecations and what they should be replaced with -->
+* Breaking change to reporting API v1alpha10.
```
llucax, line 10: Maybe it is worth mentioning the switch from `metric` to `metrics`, as it involves a few breaking changes too.

llucax (approved):
Other than that and the raising of exceptions (which I understand it might be out of the scope of this PR), LGTM.

src/frequenz/client/reporting/_types.py
cwasicki: Yes, it's on purpose to have minimal changes. There are long-standing plans of a major rework of the types once the whole microgrid/common updates are final and we go towards v1.

RELEASE_NOTES.md
cwasicki: https://github.com/frequenz-floss/frequenz-client-reporting-python/pull/241

PR #253, opened 2025-12-17
Fix aggregated values for missing data

cwasicki:
If the samples in the protobuf of the aggregated response has the value not set (e.g. when the service sends None values), the generated python code falls back to zeroes when the field is accessed. This case needs to be handled explicitly by the client, which now sets these values to NaN.


stefan-brus-frequenz (approved):
LGTM

PR #267, opened 2026-02-16
Simplify install instructions and add plotting example

Marenz:
Remove the hardcoded version from the pip install command — users should just install the latest by default.

Add a matplotlib plotting example at the end of the README showing how to visualize fetched data.

### DIFF
--- a/README.md
+++ b/README.md
@@ -200,6 +200,34 @@ df = pd.DataFrame(data)
 print(df)
 ```
 
+## Plotting data with matplotlib
+
+```bash
+pip install matplotlib
+```
+
+Using the `data` variable from any of the examples above:
+
+```python
+import matplotlib.pyplot as plt
+import matplotlib.dates as mdates
+
+timestamps = [s.timestamp for s in data]
+values = [s.value for s in data]
+
+fig, ax = plt.subplots(figsize=(14, 5))
+ax.plot(timestamps, values)
+
+# Optional: customize the plot
+ax.set_title("AC Active Power")
+ax.set_ylabel("Power (W)")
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -222,3 +220,39 @@ reporting-cli \
...
+```bash
+pip install matplotlib
+```
+
+```python
+import matplotlib.pyplot as plt
+import matplotlib.dates as mdates
+
+# Fetch data (see examples above)
+data = [
```
cwasicki: I wouldn't repeat this here but only refer to above.

README.md
```diff
@@ -222,3 +220,39 @@ reporting-cli \
     --bounds
 ```
 In addition to the default CSV format, individual samples can also be output using the `--format iter` option.
+
+## Plotting data with matplotlib
```
cwasicki, line 203: Maybe append it directly to the other python sections, i.e. before the CLI tool section.

README.md
```diff
@@ -222,3 +220,39 @@ reporting-cli \
...
+        end_time=datetime.fromisoformat("2024-05-02T00:00:00"),
+        resampling_period=timedelta(seconds=60),
+    )
+]
+
+timestamps = [s.timestamp for s in data]
+values = [s.value for s in data]
+
+fig, ax = plt.subplots(figsize=(14, 5))
+ax.plot(timestamps, values)
```
cwasicki, line 219: I would add a comment here that all of the following  customization instructions are optional.

cwasicki (dismissed):
Some optional comments.

cwasicki (approved):

PR #268, opened 2026-02-16
Fix responses with bounds

cwasicki:
Fixes a bug where bounds of zero were ignored. Also ensures that we always return a pair of upper and lower bounds, where +/-inf is returned in case the bound is not set.


micaebe (approved):
LGTM!

stefan-brus-frequenz (approved):
LGTM

